PR #21, opened 2019-05-21
bugfix/SOF-3927

mohammadidinani:


src/tools/cell.js
```diff
@@ -13,11 +13,16 @@ const ADD = math.add;
...
  * @example [[1, 0, 0], [-1, 0, 0]]
  */
-function generateTranslationCombinations(cell, point, anotherCell, amplitude = 10) {
+function generateTranslationCombinations(cell, point, anotherCell, amplitude) {
 
-    const combinations = [];
+    if (!amplitude) {
+        // Use the diagonal of the supercell as amplitude
+        const vectors = [anotherCell.vector1, anotherCell.vector2, anotherCell.vector3];
+        amplitude = parseInt(math.sqrt(vectors.reduce((a, b) => a + math.vlen(b) * math.vlen(b), 0))) + 1;
```
timurbazhirov: It's not that simple, unfortunately: 

1. You are calculating diagonal for a parallelepiped with 90 degree angles - the cell can be different. Instead we need the largest diagonal of a non-rectangular 3D version.
2. You are calculating amplitude in Angstroms which is suboptimal and gives much larger amplitude than needed - it should be divided by the smallest lattice constant

PR #27, opened 2021-03-17
Feature/sof 4212

adewyer:
hash for molecular systems


src/material.js
```diff
@@ -173,20 +173,42 @@ export class Material {
...
+    /**
+     * Gets the `inchi` string from the derived properties and returns it as the message to use for
+     * generating the hash when `isNonPeriodic` = true.
+     * @returns String
+     */
+    getNonPeriodicHashMessage() {
+        const derivedProperties = this._json.derivedProperties;
+        const inchi = lodash.isArray(derivedProperties) ? derivedProperties.find(x => x.name === "inchi") : null;
+        if (inchi) {
+            return inchi.value;
```
adewyer, line 208: Should we only use the inchi, or should we include basis information as well? I think the inchi should suffice. Adding in basis information could cause the same material to be generated with two different hashes if the basis differs. 

The only reason I could see to include the basis would be if we want a way to distinguish between conformers of the same structure, but at this stage I do not know if that is important. Conformers are something many computational chemists still struggle with handling and uniquely identifying properly, especially for acyclic structures. It's not a trivial task to say the least.

 I do not see any reason really to include the lattice.

src/material.js
```diff
@@ -173,20 +173,42 @@ export class Material {
...
+     * If `isNonPeriodic` = true
+     * The hash based on the `inchi` string for the material that is returned from getNonPeriodicHashMessage.
+     *
      * @param salt {String} Salt for hashing, empty string by default.
      * @param isScaled {Boolean} Whether to scale the lattice parameter 'a' to 1.
      */
     calculateHash(salt = '', isScaled = false) {
-        const message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
+        let message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
+        if (this._json.isNonPeriodic) {
```
timurbazhirov, line 193: `this.isNonPeriodic` would suffice

src/material.js
```diff
@@ -173,20 +173,42 @@ export class Material {
...
+            message = this.getNonPeriodicHashMessage() || message;
```
timurbazhirov, line 194: Why `II message`? We should probably raise an error if `isNonPeriodic` is set, but no inchi is supplied.

src/material.js
timurbazhirov: I agree with what you wrote above, but the implementation needs to be adjusted

src/material.js
```diff
@@ -173,20 +173,42 @@ export class Material {
...
     }
 
+    /**
+     * Gets the `inchi` string from the derived properties and returns it as the message to use for
+     * generating the hash when `isNonPeriodic` = true.
+     * @returns String
+     */
+    getNonPeriodicHashMessage() {
+        const derivedProperties = this._json.derivedProperties;
+        const inchi = lodash.isArray(derivedProperties) ? derivedProperties.find(x => x.name === "inchi") : null;
```
timurbazhirov, line 206: We are facing a conceptual problem here - `derivedProperties` and their handling seem to not be defined in made.js by design previously. We need to consider why and how the implementation needs to be adjusted.

src/material.js
```diff
@@ -106,6 +106,21 @@ export class Material {
...
+    }
+
+    /**
+     * @summary Sets the derivedProperties array based on the array pass as an argument.
+     * @param {Array} derivedProperites
+     */
+    set derivedProperties(derivedProperites) {
+        this.setProp('derivedProperties', derivedProperites)
+    }
+
```
timurbazhirov, line 123: Let's just reuse the implementation from web-app:material:

```
    getDerivedPropertyByName(name) {
        return this.getDerivedProperties().find(x => x.name === name);
    }

    static getDerivedPropertyByName(config, name) {
        const derivedProperties = lodash.get(config, 'derivedProperties', []);
        return lodash.isArray(derivedProperties) ? derivedProperties.find(x => x.name === name) : null;
    }

    getDerivedProperties(flat = false) {
        let result = this.prop('derivedProperties', []);
        if (flat) {
            result = result.map(x => {
                try {
                    return new DerivedProperty(x).flatten();
                } catch (e) {
                    return null;
                }
            }).filter(x => x !== null);
        }
        return result;
    }

```

timurbazhirov: Perhaps without the `flat` option

src/material.js
```diff
@@ -173,24 +173,54 @@ export class Material {
...
+        }
+    }
+
     get hash() {
         return this.calculateHash();
     }
 
+    set hash(hash) {
+        this.setProp('hash', hash)
+    }
```
adewyer, line 223: We may need this in the future when dealing with updating structures that have already been created.

src/material.js
```diff
@@ -173,24 +173,54 @@ export class Material {
...
      * @param salt {String} Salt for hashing, empty string by default.
      * @param isScaled {Boolean} Whether to scale the lattice parameter 'a' to 1.
      */
     calculateHash(salt = '', isScaled = false) {
-        const message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
+        let message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
+        if (this.prop('isNonPeriodic')) {
+            const nonPeriodicMessage = this.getNonPeriodicHashMessage();
+            if (nonPeriodicMessage) {
+                message = nonPeriodicMessage;
```
adewyer, line 196: return the inchi string and message so that it can be made into a hash in the same way as a periodic structure by the return function on line 201

timurbazhirov:
```js
let message;
if (!this.isNonPeriodic) {
    message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
} else {
    message = this.getNonPeriodicHashMessage();
}
```

in  `getNonPeriodicHashMessage` the error handling should be done

src/material.js
```diff
@@ -173,24 +173,54 @@ export class Material {
...
     }
 
+    /**
+     * Gets the `inchi` string from the derived properties and returns it as the message to use for
+     * generating the hash when `isNonPeriodic` = true.
+     * @returns String
+     */
+    getNonPeriodicHashMessage() {
+        const derivedProperties = this.prop('derivedProperties');
+        const inchiString = lodash.isArray(derivedProperties) ? derivedProperties.find(x => x.name === "inchi") : null;
```
timurbazhirov, line 211: Did we speak about moving the existing DerivedPropeties logic from webapp here before?

src/material.js
timurbazhirov: I'd remove this for now. The getter and setter logic is non-obvioud with `calculateHash`

src/material.js
```diff
@@ -172,25 +172,67 @@ export class Material {
...
+     * Gets the `inchi` string from the derived properties and returns it as the message to use for
+     * generating the hash when `isNonPeriodic` = true.
+     * @returns String
+     */
+    getNonPeriodicHashMessage() {
+        const inchiString = this.getDerivedProperties() ? this.getDerivedPropertyByName('inchi') : null;
+        if (inchiString) {
+            return inchiString.value;
+        } else {
+            console.error("Non-Periodic hash could not be created. Missing InChI.")
```
timurbazhirov, line 225: Should we do the actual error handling - throw an error here? See `ArrayWithIds`

src/material.js
```diff
@@ -172,25 +172,67 @@ export class Material {
...
+
+    /**
+     * @summary Returns the a specific derived property of a material, as defined by name
+     * @param {String} name
+     * @returns {Object}
+     */
+    getDerivedPropertyByName(name) {
+        return this.getDerivedProperties().find(x => x.name === name);
+    }
+
```
timurbazhirov, line 188: The above functions then should be cleaned from webapp/material https://github.com/Exabyte-io/exabyte-web-app/pull/1676/files#diff-84e19e3fddd2ab3bc58b76c9b85a23a1774ef34b16c3fdcecb9fa74fd7c61489R94

adewyer: keeping `getDerivedProperties(flat=false)` function within the web-app per our discussion of the need to flatten the materials for display on the web-app.

src/material.js
adewyer: I will test this.

src/material.js
```diff
@@ -66,6 +66,7 @@ export const defaultMaterialConfig = {
 export class Material {
     constructor(config) {
         this._json = lodash.cloneDeep(config || {});
+        this.setProp('isNonPeriodic', (config.tags || []).includes('non-periodic') || config.isNonPeriodic);
```
adewyer, line 69: moved the non-periodic property setting here from `bank_entity` in the web-app so that it only sets the value for materials rather than any bank_entity item.

src/material.js
adewyer: Still need to test this. In progress now, wanted to get SOF-Stack test pushed

src/material.js
adewyer, line 69: This works for Materials-Designer structures, but not uploaded structures.

src/material.js
```diff
@@ -172,23 +172,59 @@ export class Material {
...
+     * Gets the `inchi` string from the derived properties and returns it as the message to use for
+     * generating the hash when `isNonPeriodic` = true.
+     * @returns String
+     */
+    getNonPeriodicHashMessage() {
+        const inchiString = this.getDerivedProperties() ? this.getDerivedPropertyByName('inchi') : null;
+        if (inchiString) {
+            return inchiString.value;
+        } else {
+            throw new Error("Non-Periodic hash could not be created. Missing InChI.");
```
adewyer, line 225: updated to throw error instead of logging it

.github/workflows/cicd.yml
```diff
@@ -0,0 +1,56 @@
+name: CICD
```
adewyer, line 1: need to update epic branch w/ this file.

adewyer:
Closing PR due to errors that were easier to fix by starting a new PR where I could add one new piece of functionality for updating hashes at a time and check for breaks at each step. New PR =  [SOF-4212-ALD-2] (https://github.com/Exabyte-io/made.js/pull/47).

PR #35, opened 2021-11-15
Feature/sof 5215

adewyer:
add feature to set basis for when the non-periodic toggle switch is used.

### DIFF
--- a/src/basis/basis.js
+++ b/src/basis/basis.js
@@ -405,6 +405,9 @@ export class Basis {
                 }
             }
         }
+        if (maxDistance === 0) {
+            maxDistance = 2;
+        }
         return math.precise(maxDistance, 4);
     }
 
--- a/src/material.js
+++ b/src/material.js
@@ -150,6 +150,10 @@ export class Material {
         return this.prop('basis', undefined, true);
     }
 
+    set basis(config) {
+        return this.setProp('basis', config);
+    }
+
     // returns the instance of {ConstrainedBasis} class
     get Basis() {
         return new ConstrainedBasis({
--- a/src/material.js
+++ b/src/material.js
@@ -150,10 +150,6 @@ export class Material {
         return this.prop('basis', undefined, true);
     }
 
-    set basis(config) {
-        return this.setProp('basis', config);
-    }
-
     // returns the instance of {ConstrainedBasis} class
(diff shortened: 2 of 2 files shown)


src/material.js
```diff
@@ -152,6 +162,8 @@ export class Material {
         });
     }
 
+    set Basis(config) {return this.setProp('basis', config);}
```
timurbazhirov, line 165: We don't need this

src/material.js
```diff
@@ -144,6 +150,10 @@ export class Material {
         return this.prop('basis', undefined, true);
     }
 
+    set basis(config) {
+        return this.setProp('basis', config);
```
timurbazhirov, line 154: Shouldn't need the return

timurbazhirov (approved):

src/basis/basis.js
```diff
@@ -405,6 +405,9 @@ export class Basis {
                 }
             }
         }
+        if (maxDistance === 0) {
+            maxDistance = 2;
```
timurbazhirov, line 409: There should be some explanation of the "magic" number 2

src/basis/basis.js
```diff
@@ -405,6 +405,9 @@ export class Basis {
                 }
             }
         }
```
adewyer, line 407: I don't think we need this if statement anymore once the changes in SOF-5214 are implemented. 

Currently bumping from 0--> 2 stops the updated lattice from being set to 0. `updated lattice = max distance * lattice factor`

SOF-5214 changes this to `updatedLattice = maxDistance + lattice additive factor` so the value should never be 1 it would always be at least the value of the lattice additive factor.

src/basis/basis.js
adewyer: I don't think we need this if statement anymore once the changes in SOF-5214 are implemented. 

Currently bumping from 0--> 2 stops the updated lattice from being set to 0. `updated lattice = max distance * lattice factor`

SOF-5214 changes this to `updatedLattice = maxDistance + lattice additive factor` so the value should never be 1 it would always be at least the value of the lattice additive factor.

If you agree then I will remove these lines and update the PR.

timurbazhirov:
Closing as duplicate

PR #36, opened 2021-11-20
Feature/sof 5238

adewyer:


src/material.js
```diff
@@ -144,6 +150,10 @@ export class Material {
         return this.prop('basis', undefined, true);
     }
 
+    set basis(config) {
```
timurbazhirov, line 153: There's `setBasis` already above

adewyer: There is but it takes in 3 parameters. I want a function that will take in a basis that already exists and apply it a material.

The other function seems more appropriate if you're trying to update based on content from a text file or from changes in the basis editor.

src/material.js
```diff
@@ -158,12 +162,26 @@ export class Material {
...
+            coordinates: config.coordinates,
+            units: config.units,
+            cell: config.cell,
+            isEmpty: false
+        };
+        return new ConstrainedBasis({
+                ...newConfig
+            }
+        );
+    }
```
timurbazhirov, line 177: A setter that returns a Class instance???

src/material.js
timurbazhirov: OK

src/material.js
timurbazhirov: Let's remove this pls

src/material.js
```diff
@@ -162,20 +164,6 @@ export class Material {
...
-            units: config.units,
-            cell: config.cell,
-            isEmpty: false
-        };
-        return new ConstrainedBasis({
-                ...newConfig
-            }
-        );
-    }
-
```
adewyer: This function ended up not being needed after updating the function `convertMDMaterialToMaterial(material)` in the web-app (https://github.com/Exabyte-io/exabyte-web-app/pull/1663/files#)

src/tools/material.js
```diff
@@ -39,7 +40,28 @@ function scaleNonPeriodicLattice(material) {
     return newLattice;
 }
 
+function getBasisConfigTranslatedToCenter(material) {
```
timurbazhirov, line 43: The idea is that the tools are working on a material object and return a modified material. We can extract the config by running `toJSON` on `material.Basis`

src/tools/material.js
```diff
@@ -32,7 +34,26 @@ function scaleLatticeToMakeNonPeriodic(material) {
...
+ * so that the center of the material and lattice are aligned.
+ * @param material {Material}
+ * */
+function getBasisConfigTranslatedToCenter(material) {
+    const lattice = new Lattice(material.lattice);
+    const basis = new Basis({
+        ...material.Basis.toJSON()
+    });
+    basis.toCartesian();
+    const centerOfCoordinates = basis.centerOfCoordinatesPoint;
```
timurbazhirov, line 48: Seems like too much code for what it needs to be

Why not just work on the `material.Basis` directly? There's also `material.toCartesian` function used above in the file...

PR #40, opened 2021-12-06
Feature/sof 5215

adewyer:
Function for Lattice update for non-periodic structures added to tools/materials.js

### DIFF
--- a/src/tools/material.js
+++ b/src/tools/material.js
@@ -1,3 +1,5 @@
+import {Basis} from "../basis/basis";
+import {Lattice} from "../lattice/lattice";
 /**
  * Scales one lattice vector for the given material
  * @param material {Material} The material acted upon.
@@ -18,6 +20,26 @@ function scaleOneLatticeVector(material, key = 'a', factor = 1.0) {
 
 }
 
+function scaleNonPeriodicLattice(material) {
+    const lattice = new Lattice(material.lattice);
+    const basis = new Basis({
+        ...material.basis,
+        cell: lattice.vectorArrays,
+        units: "crystal"
+    })
+    basis.toCartesian();
+    basis.toJSON();
+    //TODO: Change to match the new scaling factor variable from SOF-5214
+    const scalingFactor = 2.0;
+    const newLattice = new Lattice({
+        a: basis.maxPairwiseDistance * scalingFactor,
+        type: "CUB"
+    });
+    newLattice.toJSON();
+    return newLattice;
+}
+
 export default {
     scaleOneLatticeVector,
+    scaleNonPeriodicLattice,
 }
--- a/src/tools/material.js
+++ b/src/tools/material.js
@@ -25,24 +25,20 @@ function scaleOneLatticeVector(material, key = 'a', factor = 1.0) {
  * Returns the lattice for a non-periodic structure. The lattice has been scaled to accomodate the size of the
  * non-periodic structure.
  * @param material {Material}
- * @returns lattice
+ * @returns {Object}
  */
 function scaleLatticeToMakeNonPeriodic(material) {
-    const lattice = new Lattice(material.lattice);
     const basis = new Basis({
-        ...material.basis,
-        cell: lattice.vectorArrays,
-        units: "crystal"
+        ...material.Basis.toJSON(),
     })
     basis.toCartesian();
-    basis.toJSON();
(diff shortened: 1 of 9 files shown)


adewyer:
Do we want a test for this new function? I was not sure if a test was needed as the other option inside the tools/materials.js file does not have a test associated with it..

src/tools/material.js
```diff
@@ -18,6 +20,26 @@ function scaleOneLatticeVector(material, key = 'a', factor = 1.0) {
...
+function scaleNonPeriodicLattice(material) {
+    const lattice = new Lattice(material.lattice);
+    const basis = new Basis({
+        ...material.basis,
+        cell: lattice.vectorArrays,
+        units: "crystal"
+    })
+    basis.toCartesian();
+    basis.toJSON();
+    //TODO: Change to match the new scaling factor variable from SOF-5214
```
adewyer, line 32: I need to merge the code from SOF-5214 before I can update this.

timurbazhirov: Then let's start this branch from code that includes it already??

src/tools/material.js
```diff
@@ -18,6 +20,26 @@ function scaleOneLatticeVector(material, key = 'a', factor = 1.0) {
 
 }
 
+function scaleNonPeriodicLattice(material) {
```
timurbazhirov, line 23: scaleLatticeToMakeNonPeriodic

src/tools/material.js
```diff
@@ -18,6 +21,27 @@ function scaleOneLatticeVector(material, key = 'a', factor = 1.0) {
...
+    })
+    basis.toCartesian();
+    const scalingFactor = nonPeriodicLatticeScalingFactor;
+    const newLattice = new Lattice({
+        a: basis.maxPairwiseDistance * scalingFactor,
+        type: "CUB"
+    });
+
+    return newLattice.toJSON();
+}
```
timurbazhirov, line 42: (1) I'd just implement the function to be modifying material in-place like the `scaleOneLatticeVector` above.
(2) why do we need to convert basis to cartesian? the `maxPairwiseDistance` should be taking care of that already on its own. I think the distance only makes sense in cartesian coordinates.
(3) the implementation should just be

```js
function scaleLatticeToMakeNonPeriodic(material) {
    material.lattice = new Lattice({
        a: material.Basis.maxPairwiseDistance * nonPeriodicLatticeScalingFactor,
        type: "CUB"
    });
}
```

src/tools/material.js
```diff
@@ -18,6 +20,19 @@ function scaleOneLatticeVector(material, key = 'a', factor = 1.0) {
...
 }
 
+/**
+ * Updates the size of a materials lattice using the minimumLatticeSize function.
+ * The new size of the material is calculated based on the materials basis.
+ * @param material {Material}
+ */
+function scaleLatticeToMakeNonPeriodic(material) {
+    material.lattice = new Lattice({
+        a: material.Basis.minimumLatticeSize,
```
adewyer, line 30: I realized i did not update this function to the proper one. It should call `minimumLatticeSize` function, which was implemented for updating the lattice in the `hooks.js` function for the web-app. It should also be used here. https://github.com/Exabyte-io/made.js/blob/feature/SOF-5215-ALD/src/basis/basis.js#:~:text=*/-,get%20minimumLatticeSize()%20%7B,%7D,-/**

src/tools/material.js
```diff
@@ -18,6 +20,19 @@ function scaleOneLatticeVector(material, key = 'a', factor = 1.0) {
...
 }
 
+/**
+ * Updates the size of a materials lattice using the minimumLatticeSize function.
+ * The new size of the material is calculated based on the materials basis.
+ * @param material {Material}
+ */
+function scaleLatticeToMakeNonPeriodic(material) {
+    material.lattice = new Lattice({
+        a: material.Basis.getMinimumLatticeSize(),
```
adewyer, line 30: I realized i did not update this function to the proper one. It should call getMinimumLatticeSize function, which was implemented for updating the lattice in the hooks.js function for the web-app. It should also be used here.https://github.com/Exabyte-io/made.js/blob/feature/SOF-5215-ALD/src/basis/basis.js#:~:text=*/-,getMinimumLatticeSize(latticeScalingFactor%20%3D%20nonPeriodicLatticeScalingFactor)%20%7B,%7D,-/**

PR #41, opened 2021-12-08
Feature/SOF 4398

tjduigna:


src/basis/basis.js
```diff
@@ -106,53 +114,79 @@ export class Basis {
...
+        return this._coordinates.getArrayElementByIndex(idx);
+    }
 
-    get elements() {return this._elements.toJSON()}
+    get elements() {
+        return this._elements.toJSON();
+    }
 
-    get elementsArray() {return this._elements.array}
+    get elementsArray() {
```
timurbazhirov, line 143: Should we keep the one-line functions?

src/lattice/unit_cell.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    vectorC() {
+        return [this.cx, this.cy, this.cz];
+    }
+
+    axes() {
+        return [this.vectorA(), this.vectorB(), this.vectorC()];
+    }
+}
```
timurbazhirov, line 31: Looks like the whole file is being added anew - are you sure?

tjduigna: This was the approach radik took. There is a TODO to remove this class and a TODO to remove reference to it in the only location it's used. We need to know if it can be safely removed before actually removing it.

src/basis/constrained_basis.js
```diff
@@ -71,15 +78,14 @@ export class ConstrainedBasis extends Basis {
...
             const element = entry[0];
             const coordinate = entry[1];
             const constraint = entry[2];
-            return s.sprintf('%-4s', element) +
-                coordinate.map(x => s.sprintf('%14.9f', x).trim()).join(' ') +
-                ' ' + constraint.map(x => x ? 1 : 0).join(' ');
+            return `${
+                s.sprintf("%-4s", element) +
+                coordinate.map((x) => s.sprintf("%14.9f", x).trim()).join(" ")
+            } ${constraint.map((x) => (x ? 1 : 0)).join(" ")}`;
```
timurbazhirov, line 88: Let's discuss and finalize the use of string substitution everywhere

src/lattice/reciprocal/symmetry_points.js
```diff
@@ -674,765 +674,773 @@ const POINTS = {
...
 
     [LATTICE_TYPE.MCLC]: ({a, b, c, alpha, beta, gamma}) => {
         const cosAlpha = Math.cos(alpha / 180 * Math.PI);
+        let e;
+        let n;
+        let p;
+        let f;
+        let m;
+        let d;
+        let v;
```
timurbazhirov, line 1013: maybe let's allow multiple definitions on one line `let e, n, p, f, g`?

src/basis/basis.js
```diff
@@ -309,16 +351,19 @@ export class Basis {
...
         const clsInstance = new Basis(this.toJSON());
         clsInstance.toStandardRepresentation();
-        return clsInstance.elementsAndCoordinatesArray.map((entry) => {
-            const element = entry[0];
-            const coordinate = entry[1];
-            const toleratedCoordinates = coordinate.map(x => math.round(x, HASH_TOLERANCE));
-            return `${element} ${toleratedCoordinates.join()}`;
-        }).sort().join(';') + ";";
+        return `${clsInstance.elementsAndCoordinatesArray
+            .map((entry) => {
```
timurbazhirov, line 359: Let's try and keep the first function call on the same line as the object reference?

tjduigna: This closed issue https://github.com/eslint/eslint/issues/10977 almost discusses what we want. In general, I'm not sure it's possible to force in-lining the first in a chain of method calls. For this particular example, it can still be refactored for readability.

src/basis/basis.js
tjduigna: Prettier abhors one-liners but eslint allows it through a `brace-style` rule option. In order for prettier to not clash with eslint (with the rule set), we need to individually ignore files containing one-liners. IMO the use of such an "opinionated" formatter is only beneficial if we release control of what we deem as "stylish" in favor of the ruthless convention.

src/basis/basis.js
timurbazhirov: Let's see if we can use the Airbnb example config from https://github.com/airbnb/javascript/tree/master/linters to support example 9.3 in https://github.com/airbnb/javascript/tree/master/linters

tjduigna: I think this example is a pathological case due to the string interpolation. in `get atomicPositions` right below it it contains the style we prefer but of course it's not chained. Then in `hasEquivalentCellTo` as a chained call it is inconsistent again. The only way around it would be to sprinkle `// prettier-ignore` comments or remove `prettier` entirely and leave considerably style formatting up to the developer, with `eslint` doing what it can.

timurbazhirov (approved):

PR #43, opened 2021-12-09
Feature/sof 4398 tb

timurbazhirov:


tjduigna:
I figured it out. the `lint-staged` directive in `package.json` was including `*.md` in the `prettier` call prior to commit so the README was getting auto-formatted.. not sure why we don't want a pretty README 😄  but it's resolved.

PR #47, opened 2021-12-18
Feature/sof 4212 ald 2

adewyer:


src/material.js
```diff
@@ -172,6 +172,19 @@ export class Material {
         return new Lattice(this.lattice);
     }
 
+    getDerivedPropertiesForHash() {
```
adewyer, line 175: Moving the derivedProperty functions from the materials class in the web-app to made-js caused errors in hash generation due to initial materials being generated containing no derived properties.

src/material.js
```diff
@@ -172,6 +172,19 @@ export class Material {
...
+        const derivedProperties = this._json.derivedProperties;
+        if (derivedProperties) {
```
adewyer, line 177: We only want to run the code below if `derivedProperties` has been created and filled with properties. Otherwise the Error will get thrown because an empty array leads to `!inchi`, which triggers the error.

src/material.js
```diff
@@ -174,6 +191,19 @@ export class Material {
         return new Lattice(this.lattice);
     }
 
+    getDerivedPropertiesForHash() {
```
timurbazhirov, line 194: The name of this function is still incorrect - we never return derivedProperties, only a string for the inchi value

src/material.js
```diff
@@ -174,6 +191,19 @@ export class Material {
...
+        const derivedProperties = this.getDerivedProperties();
+        if (derivedProperties.length > 0) {
+            const inchi = lodash.isArray(derivedProperties) ? this.getDerivedPropertyByName('inchi') : null;
```
timurbazhirov, line 197: No need for the `isArray` check anymore b/c of line 122

src/material.js
```diff
@@ -184,8 +214,14 @@ export class Material {
...
-    calculateHash(salt = "", isScaled = false) {
-        const message = `${this.Basis.hashString}#${this.Lattice.getHashString(isScaled)}#${salt}`;
+    calculateHash(salt = '', isScaled = false) {
+        let message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
+        if (this.isNonPeriodic) {
+            const inchiString = this.getDerivedPropertiesForHash();
+            if (inchiString) {
+                message = inchiString;
+            }
+        }
```
timurbazhirov, line 224: This logic is masking the problem - if we want to extract the inchi hash, but it's not available, then the hash is still calculated as for the periodic materials. That should not happen, we should throw error.

adewyer: The error check is implemented above https://github.com/Exabyte-io/made.js/pull/47/files#:~:text=Show%20conversation-,if%20(inchi)%20%7B,%7D,-return%20null.

If we try the error here it fails during non-periodic materials creation with the MD because the derived properties have not been created yet.

src/material.js
```diff
@@ -174,6 +191,19 @@ export class Material {
         return new Lattice(this.lattice);
     }
 
+    getInchiStringForHash() {
+        const derivedProperties = this.getDerivedProperties();
+        if (derivedProperties.length > 0) {
+            const inchi = derivedProperties ? this.getDerivedPropertyByName('inchi') : null;
```
timurbazhirov, line 197: There's no need for this check. The default value is an empty array which is a truth value

src/material.js
```diff
@@ -174,6 +191,19 @@ export class Material {
...
+            if (inchi) {
+                return inchi.value
+            } else {
+                throw new Error("Hash cannot be created. Missing InChI string in derivedProperties")
```
timurbazhirov, line 201: Let's move this to the end of the checks if inchi cannot be extracted, we should not be calling this function

src/material.js
```diff
@@ -184,13 +214,26 @@ export class Material {
...
-        const message = `${this.Basis.hashString}#${this.Lattice.getHashString(isScaled)}#${salt}`;
+    calculateHash(salt = '', isScaled = false) {
+        let message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
+        if (this.isNonPeriodic) {
+            const inchiString = this.getInchiStringForHash();
+            if (inchiString) {
+                message = inchiString;
+            }
+            else {
+                throw new Error("Hash cannot be created. Missing InChI string");
```
timurbazhirov, line 225: No need to check in 2 places

```
let message;
if (this.isNonPeriodic) {
    message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
} else {
    message = this.getInchiStringForHash()
}
```

src/material.js
```diff
@@ -77,10 +77,15 @@ export class Material {
...
             name: this.name || this.formula,
             isNonPeriodic: this.isNonPeriodic || false,
+            hash: this.hash,
         };
     }
 
     clone(extraContext) {
+        const resetHash = extraContext ? ("resetHash" in extraContext) : false;
+        if (resetHash) {
+            this.setProp('hash', '');
```
timurbazhirov, line 87: This will reset the has for the current material too, we don't want that.

Moreover, `extraContext` already allows us to reset hash by just passing: `{hash: ""}` when invoking. Nothing else.

timurbazhirov (approved):

timurbazhirov (approved):

PR #53, opened 2022-01-13
SOF-5460

adewyer:


src/material.js
```diff
@@ -111,15 +111,15 @@ export class Material {
      * @returns {Object}
      */
     getDerivedPropertyByName(name) {
-        return this.getDerivedProperties().find(x => x.name === name);
+        return this.getDerivedProperties().find((x) => x.name === name);
```
adewyer, line 114: I think the commit check that has been implemented changed this.

src/material.js
```diff
@@ -215,29 +214,45 @@ export class Material {
...
+     * @param {Boolean} isScaled
+     * returns {String}
+     */
+    calculateScaledHash(salt = "", isScaled = true) {
+        return this.calculateHash(salt, isScaled);
+    }
+
     set hash(hash) {
-        this.setProp('hash', hash);
+        this.setProp("hash", hash);
```
timurbazhirov, line 240: There was a reason why we had the previous logic that avoided directly storing the hash and recalculated it "on-the-fly" from lattice/basis - it was to avoid a "corrupted" state where the lattice and basis are changed but a hash is not.

That's why we didn't set the `hash` in a constructor (https://github.com/Exabyte-io/made.js/pull/53/files#diff-c664f5e378394d9c3fc8b008011f236c1beb1a8f1a2f741e86dd99712b970a32R59) and toJSON in https://github.com/Exabyte-io/made.js/pull/53/files#diff-c664f5e378394d9c3fc8b008011f236c1beb1a8f1a2f741e86dd99712b970a32R74 does not return it either. Hash is a "second-tier" variable.

When introducing inchi, we have to understand how this logic can be adjusted.

One way would be to keep the inchi value inside the basis to make sure they both are always updated together.

Otherwise, we have to understand how to assert a correct state of basis vs hash relationship at all times with some other checks.

src/material.js
```diff
@@ -215,24 +214,17 @@ export class Material {
...
-            message = this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
+            message =
+                this.Basis.hashString + "#" + this.Lattice.getHashString(isScaled) + "#" + salt;
         } else {
             message = this.getInchiStringForHash();
         }
         return CryptoJS.MD5(message).toString();
     }
 
-    set hash(hash) {
```
adewyer, line 228: if we are not storing this information then we shouldn't need these two functions.

src/material.js
```diff
@@ -215,31 +214,31 @@ export class Material {
      * @param salt {String} Salt for hashing, empty string by default.
      * @param isScaled {Boolean} Whether to scale the lattice parameter 'a' to 1.
      */
-    calculateHash(salt = '', isScaled = false) {
```
adewyer, line 218: reverted function back to original. Moved chemicalHash creation to function on line 235

PR #55, opened 2022-01-21
added in function to check number of atoms in xyz text

adewyer:


src/basis/basis.js
```diff
@@ -7,6 +7,12 @@ import { ATOMIC_COORD_UNITS, HASH_TOLERANCE } from "../constants";
 import { Lattice, nonPeriodicLatticeScalingFactor } from "../lattice/lattice";
 import math from "../math";
 
+
+/**
+ * Maximum number of atoms allowed in materials designer basis or a file upload basis.
```
timurbazhirov, line 12: This message is against the encapsulation of concepts. A person using this library has to know about materials designer and file upload to understand this. A better approach would be to call this something independent of the other packages, like `maximumNumberAtomsSmallMolecule`. Then explain the logic that everything beyond that is to be considered a "large" molecule and thus subject to different processing.

timurbazhirov: Alternatively, we can just include `nonPeriodicMaxAtomCount` inside the `Basis` class as a static variable

adewyer: changed to static variable.

src/parsers/xyz.js
```diff
@@ -33,16 +33,41 @@ function validateLine(xyzLine, index) {
...
-        .trim()
-        .lines()
-        .filter((x) => x.trim() !== "")
-        .forEach(validateLine);
+    try {
+        s(xyzTxt)
+            .trim()
+            .lines()
+            .filter((x) => x.trim() !== "")
+            .forEach(validateLine);
```
timurbazhirov, line 61: Let's rename previous `validate` to `validateFormat` and then use it in lines 57-61 inside `validateAll`

tests/parsers/xyz.js
```diff
@@ -11,4 +14,12 @@ describe("Parsers:XYZ", () => {
             "units",
         ]);
     });
+    it("should return error number for a structure having less than the maximum number of atoms", () => {
+        const error = validate(smallBasis);
+        expect(error).to.be.equal(0);
+    });
+    it("should return the error number for a structure having more than the maximum number of atoms", () => {
+        const error = validate(largeBasis);
+        expect(error).to.be.equal(2001);
```
timurbazhirov, line 23: Should also check for 1001 error

src/parsers/xyz.js
```diff
@@ -34,15 +34,52 @@ function validateLine(xyzLine, index) {
...
+    } catch (err) {
+        return 1001;
+    }
+    return 0;
```
timurbazhirov, line 66: The logic for dealing with error codes should be in validateAll.

This function should just throw an Error like validateNumberOfAtoms above

timurbazhirov (approved):

PR #56, opened 2022-01-26
Feature/sof 5478

adewyer:


src/tools/material.js
```diff
@@ -27,6 +26,7 @@ function scaleOneLatticeVector(material, key = "a", factor = 1.0) {
  * @param material {Material}
  */
 function scaleLatticeToMakeNonPeriodic(material) {
+    material.toCartesian();
```
timurbazhirov, line 29: We should add a couple of tests to assert situation arising in this task doesn't happen in the future:

1. Repeatedly apply `getBasisConfigTranslatedToCenter` and assert nothing happens second time
2. Same for scaleLatticeToMakeNonPeriodic
3. Repeatedly apply `scaleLatticeToMakeNonPeriodic(getBasisConfigTranslatedToCenter ())` and assert nothing happens the second time

timurbazhirov (approved):

adewyer:
Code base will need to be tested w/ all other merges in the epic.

PR #64, opened 2022-08-31
SOF-6252: exabyte peer dependencies

k0stik:


timurbazhirov (approved):

src/material.js
```diff
@@ -78,6 +78,10 @@ export class Material extends HasMetadataNamedDefaultableInMemoryEntity {
         return super.name || this.formula;
     }
 
+    set name(name) {
+        super.name = name;
+    }
+
```
timurbazhirov, line 84: To avoid similar problems in the future - should we define `alternativeName` getter in code.js and use it inside name:

in code.js named entity:
```
get alternativeName() {throw "Alternative name not defined"}

get name () {return this.prop("name") || this.alternativeName || ''}
```

in made.js/material, instead of redefining name getter:

`get alternativeName() {return this.formula}`

PR #79, opened 2023-05-25
feature/SOF-6597-1

VsevolodX:
Mock conversion function, only JSON is handled currently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules/
 .idea/**
 .husky
 .nyc_output
+.DS_Store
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules/
 .idea/**
 .husky
 .nyc_output
+.DS_Store
```
unsigned6, line 7: Better configure it's globally once in your system:
https://dev.to/lpheller/quickly-ignore-dsstore-files-globally-ecd

PR #80, opened 2023-05-26
feature/SOF-6597-2

VsevolodX:
Handling of JSON and POSCAR files. Although when parsing POSCAR lattice type is set to default ("TRI") now, no automatic detection.


.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules/
 .idea/**
 .husky
 .nyc_output
+.DS_Store
```
unsigned6, line 7: https://dev.to/lpheller/quickly-ignore-dsstore-files-globally-ecd

unsigned6 (commented):
Looks good

src/parsers/native_formats.js
```diff
@@ -0,0 +1,52 @@
...
+            return Poscar.fromPoscar(text);
+        case NATIVE_FORMAT.UNKNOWN:
+            throw new Error(`Unknown format`);
+        // TODO:  add more formats
+        default:
+            throw new Error(`Unsupported format: ${format}`);
+    }
+}
+
+export default {
```
k0stik, line 49: You already use naming export for `export const NATIVE_FORMAT`
Let's have a naming export for `detectFormat` and `convertFromNative` then:
`export function detectFormat() ...`
`export function detectFormat() ...`

timurbazhirov: To be using default export

timurbazhirov (approved):

PR #81, opened 2023-06-07
feature/SOF-6598

VsevolodX:
First working draft. Graphene with ibrav=0 and Ni with ibrav=1 convert correctly. I will re-check lattice detection algorithm, add fixtures with different units and ibravs


src/parsers/native_format_parsers.js
```diff
@@ -1,10 +1,11 @@
+import Espresso from "./espresso";
 import Poscar from "./poscar";
 
 const NATIVE_FORMAT = {
     JSON: "json",
     POSCAR: "poscar",
     CIF: "cif",
-    PWX: "pwx",
+    IN: "in",
```
timurbazhirov, line 8: Probably better to call this `QE`, maybe?

VsevolodX: I thought in terms of file extension. .json, .cif, .in

src/parsers/espresso.js
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+    alat = 1; // TODO: remove and change this
+    return {
+        cell: cellVectors,
+        alat,
+        units,
+    };
+}
+
+/**
+ * @summary Creates cell from ibrav and celldm(i) parameters
```
timurbazhirov, line 156: 1. Let's include a link to the QE documentation where this is explained pls
2. We should include tests for this function specifically, it's pretty complicated
3. It's not a huge deal for this specific case, but in JS we use `camelCase` not `snake_case` - see how we deal with this in reciprocal cell symmetry points definitions pls.

src/parsers/espresso.js
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+            cosab = system.celldm[6] || 0.0;
+        }
+    } else if (system.a) {
+        // a, b, c, cosAB, cosAC, cosBC in Angstrom
+        throw new Error("params_to_cell() does not yet support A/B/C/cosAB/cosAC/cosBC");
+    } else {
+        throw new Error("Missing celldm(1)");
+    }
+
+    switch (system.ibrav) {
```
timurbazhirov, line 201: I think the real mapping here should be just for the ibrav numbers and lattice types, but the cell definition for lattice types should be inside Lattice, or Cell, perhaps.

src/parsers/espresso.js
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+    const regex = /&CONTROL|&SYSTEM|CELL_PARAMETERS|ATOMIC_POSITIONS/i;
+    return regex.test(fileContent);
+}
+
+/**
+ * @summary Function to convert Fortran namelist to JavaScript object.
+ * @param {String} data - Fortran string.
+ * @return {Object}
+ */
+function parseFortranNameList(data) {
```
timurbazhirov, line 48: This should go to `parsers.fortran`, perhaps?

timurbazhirov: Maybe even to `code.js:utils` since this can be used elsewhere - by wode.js, for instance, later on

src/parsers/espresso.js
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+    return result;
+}
+
+/**
+ * @summary Read unit cell parameters from CELL_PARAMETERS card
+ * @param {String} cards
+ * @param {Number} alat
+ * @returns {cell: Number[][], alat: Number, units: String}
+ */
+function getCellParameters(cards, alat = null) {
```
timurbazhirov, line 101: This function should be tested with fixtures as well

src/parsers/espresso.js
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+    return { cell, alat, units, type };
+}
+
+/**
+ * @summary function to get atomic positions from the ATOMIC_POSITIONS card
+ * @param {String} cards
+ * @param {Number} nAtoms
+ * @returns {[String], [Number], [Boolean], String}
+ */
+function getAtomicPositions(cards) {
```
timurbazhirov, line 381: Same as above - should be tested separately

src/parsers/espresso.js
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+        c: cell.cell[2],
+        alat: cell.alat,
+        units: cell.units,
+    });
+    if (cell.type) lattice.type = cell.type;
+
+    const basis = new ConstrainedBasis({
+        elements,
+        coordinates,
+        units,
```
timurbazhirov, line 479: One of `ATOMIC_COORD_UNITS.crystal` or `ATOMIC_COORD_UNITS.cartesian`

src/parsers/espresso.js
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+        case 5:
+            tx = math.sqrt((1.0 - cosab) / 2.0);
+            ty = math.sqrt((1.0 - cosab) / 6.0);
+            tz = math.sqrt((1 + 2 * cosab) / 3.0);
+            cell = [
+                [tx * alat, -ty * alat, tz * alat],
+                [0, 2 * ty * alat, tz * alat],
+                [-tx * alat, -ty * alat, tz * alat],
+            ];
+            type = LATTICE_TYPE.TRI;
```
timurbazhirov, line 251: Should not be TRI

src/parsers/espresso.js
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+    return {
+        cell: cellVectors,
+        alat,
+        units,
+    };
+}
+
+/**
+ * @summary Creates cell from ibrav and celldm(i) parameters
+ * @param {Object} system
```
timurbazhirov, line 157: What's inside system?

src/parsers/espresso.js
```diff
@@ -36,75 +37,70 @@ function toEspressoFormat(materialOrConfig) {
...
+//                 }
+//             }
+//             result[blockName.slice(1)] = blockContent;
+//         } else {
+//             result.cards = block;
+//         }
+//     });
+//
+//     return result;
+// }
```
timurbazhirov, line 94: Should not be committed

src/parsers/fortran_parser.js
```diff
@@ -0,0 +1,180 @@
...
+    });
+    return namelists;
+}
+
+/**
+ * @summary Parses Fortran namelists and cards data from a string for a QE input file
+ * @param {String} text
+ * @returns {Object}
+ */
+export function parseFortranFile(text) {
```
timurbazhirov, line 127: This is specific to QE only, should be called that and moved out of this file

src/parsers/espresso.js
```diff
@@ -25,6 +29,273 @@ function toEspressoFormat(materialOrConfig) {
...
+            break;
+        case 14:
+            type = LATTICE_TYPE.TRI;
+            break;
+        default:
+            throw new Error(`ibrav = ${system.ibrav} not implemented`);
+    }
+
+    let vectors = []; // TODO: implement
+    vectors = Lattice.vectorsFromType(type, a, b, c, cosab, cosac, cosbc);
```
VsevolodX, line 228: @timurbazhirov Here the function will receive all parameters converted in accordance with QE specs, while inside of it in ` lattice_bravais` math will be as in https://arxiv.org/pdf/1004.2974.pdf

src/lattice/lattice_bravais.js
```diff
@@ -122,6 +122,152 @@ export class LatticeBravais {
...
+     * @param {number} a
+     * @param {number} [b]
+     * @param {number} [c]
+     * @param {number} [alpha] - angle between b and c in radians
+     * @param {number} [beta] - angle between a and c in radians
+     * @param {number} [gamma] - angle between a and b in radians
+     * @returns {{vectors: number[][], type: string}}
+     */
+    // eslint-disable-next-line no-unused-vars
+    static vectorsFromType(type, a, b, c, alpha, beta, gamma) {
```
timurbazhirov, line 150: Pass object instead

tests/parsers/fortran_parser.js
```diff
@@ -0,0 +1,16 @@
+import { expect } from "chai";
```
timurbazhirov, line 1: To move to code.js:

- create `utils/parsers` folder
- create `utils/parsers/settings.js` for regex definitions
- move to `utils/parsers/fortran.js` for the function itself

src/parsers/utils.js
```diff
@@ -0,0 +1,58 @@
...
+const fortranNamelistRegex = "&({0})((?:\\s|\\S)*?)\\/"; // capturing group: 1 - namelist name, 2 - data
+const fortranCardsRegex = "^\\s*\\/(?![\\s\\S]*^\\/)([\\s\\S]*)"; // capturing group: 1 cards section string
+const keyValueRegex = "^\\s*{0}\\s*=\\s*{1}\\s*\\n"; // 0 = string value name, 1 = str|bool|number
+const stringRegex = "([+\\w.\\-\\/]*)"; // capturing group: 1 - string value
+const fortranStringRegex = "'([\\w.\\-\\+\\/ ]*)'"; // capturing group: 1 - string value
+const fortranArrayRegex = "^\\s*{0}\\({1}\\)\\s*=\\s*{2}\\s*\\n"; // /gm, capturing groups: 0 - array name, 1 - array index, 2 - value
+const fortranBooleanRegex = "\\.(true|false)\\."; // capturing group: 1 - boolean value
+
+/**
+ * @summary Formats a string with other string to be used as a regex.
```
timurbazhirov, line 11: Rephrase to understand better

src/parsers/utils.js
```diff
@@ -0,0 +1,58 @@
+const fortranDoubleRegex = "([-+]?\\d*\\.?\\d*(?:[Eed][+-]?\\d+)?)";
+const fortranNamelistRegex = "&({0})((?:\\s|\\S)*?)\\/"; // capturing group: 1 - namelist name, 2 - data
+const fortranCardsRegex = "^\\s*\\/(?![\\s\\S]*^\\/)([\\s\\S]*)"; // capturing group: 1 cards section string
+const keyValueRegex = "^\\s*{0}\\s*=\\s*{1}\\s*\\n"; // 0 = string value name, 1 = str|bool|number
+const stringRegex = "([+\\w.\\-\\/]*)"; // capturing group: 1 - string value
+const fortranStringRegex = "'([\\w.\\-\\+\\/ ]*)'"; // capturing group: 1 - string value
+const fortranArrayRegex = "^\\s*{0}\\({1}\\)\\s*=\\s*{2}\\s*\\n"; // /gm, capturing groups: 0 - array name, 1 - array index, 2 - value
+const fortranBooleanRegex = "\\.(true|false)\\."; // capturing group: 1 - boolean value
```
timurbazhirov, line 8: Fortran-specific regex should be isolated

src/parsers/espresso.js
```diff
@@ -25,6 +31,320 @@ function toEspressoFormat(materialOrConfig) {
...
+    return espressoRegex.test(fileContent);
+}
+
+/**
+ * @summary Read unit cell parameters from CELL_PARAMETERS card
+ * @param {String} text - cards data
+ * @param {Number} alat - lattice parameter
+ * @returns {{vectors: number[][], alat: Number, units: string}}
+ */
+function getCellParameters(text, alat = null) {
```
timurbazhirov, line 50: getCellConfig

src/parsers/espresso.js
```diff
@@ -1,7 +1,12 @@
 import _ from "underscore";
 import s from "underscore.string";
 
+import { ConstrainedBasis } from "../basis/constrained_basis";
```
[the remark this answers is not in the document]
spectre007: Ideally, we should have parsers for each version of QE. I don't expect too many variations between versions, but it's important to catch differences if there are some.

I suggest we implement a collection of these parsers where applicable, for instance for version 5.2 and for version 7.1. However, these functions should not be called directly but through an interface that resolves which function should be used (for a given version number). These parsing functions also need to be tested individually, i.e., we need fixtures for each of the versions as well (similar to [Jmol](https://sourceforge.net/p/jmol/code/HEAD/tree/trunk/Jmol-datafiles/quantumEspresso/)).
Since this will undoubtedly grow the lines of code for parsers in this repo, we may also want to move it to a separate package entirely.

These are just my initial thoughts, I'd be happy to discuss this further together.

src/lattice/lattice_bravais.js
```diff
@@ -122,6 +122,155 @@ export class LatticeBravais {
...
+     * @param {Number} config.a
+     * @param {Number} config.b
+     * @param {Number} config.c
+     * @param {Number} config.alpha - angle between b and c in degrees
+     * @param {Number} config.beta - angle between a and c in degrees
+     * @param {Number} config.gamma - angle between a and b in degrees
+     * @returns {{vectors: Number[][], type: String}}
+     */
+    // eslint-disable-next-line no-unused-vars
+    static vectorsFromType({ type, a = 1, b = a, c = a, alpha = 90, beta = 90, gamma = 90 }) {
```
timurbazhirov, line 151: Remove. This is already done in `fromBravais`, see `PRIMITIVE_CELLS`

src/parsers/espresso.js
```diff
@@ -25,6 +32,335 @@ function toEspressoFormat(materialOrConfig) {
...
+    let _b = celldm ? celldm[2] * celldm[1] : b; // celldm[2] is b/a
+    let _c = celldm ? celldm[3] * celldm[1] : c; // celldm[3] is c/a
+
+    if (celldm && a) {
+        throw new Error("Both celldm and A are given");
+    } else if (celldm) {
+        // celldm(x) in bohr - from QE docs
+        _a *= coefficients.BOHR_TO_ANGSTROM;
+        _b *= coefficients.BOHR_TO_ANGSTROM;
+        _c *= coefficients.BOHR_TO_ANGSTROM;
```
timurbazhirov, line 200:
```
[ _a, _b, _c] = [ _a, _b, _c].map(x => x*coefficients.BOHR_TO_ANGSTROM)
```

src/parsers/espresso.js
```diff
@@ -25,6 +32,335 @@ function toEspressoFormat(materialOrConfig) {
...
+            if (celldm) {
+                alpha = math.acos(celldm[4]);
+                beta = math.acos(celldm[5]);
+                gamma = math.acos(celldm[6]);
+            }
+            break;
+        default:
+            throw new Error(`ibrav = ${system.ibrav} not implemented`);
+    }
+    const typeExtended = "Not implemented";
```
timurbazhirov, line 283: Isolate the switch

src/parsers/fortran_parser.js
```diff
@@ -0,0 +1,101 @@
...
+    } catch (err) {
+        throw new Error("Incorrect fortran file");
+    }
+
+    if (!output) {
+        throw new Error("No namelists found");
+    }
+
+    const match = text.match(regex.cards);
+    const qeCardsRegex = /ATOMIC_SPECIES|ATOMIC_POSITIONS/;
```
timurbazhirov, line 92: These should be `enums`:

PWX_ENUMS = {
    atomicSpeciesCard: {
        header: "ATOMIC_SPECIES"
    }
}
```

File: src/parsers/espresso.js (hunk 2 above)
Comment (VsevolodX) [reply src/parsers/espresso.js:172]: 1. ok
2. tests will be in follow up task
3. I don't see a snake_case usage here

File: src/parsers/espresso.js (hunk 8 above)
Comment (VsevolodX) [reply src/parsers/espresso.js]: Fixed

File: src/parsers/espresso.js [hunk 23]
```diff
@@ -25,6 +30,468 @@ function toEspressoFormat(materialOrConfig) {
...
+            tz = math.sqrt((1 + 2 * cosab) / 3.0);
+            a_prime = alat / math.sqrt(3);
+            u = tz - 2 * math.sqrt(2) * ty;
+            v = tz + math.sqrt(2) * ty;
+            cell = [
+                [u * a_prime, v * a_prime, v * a_prime],
+                [v * a_prime, u * a_prime, v * a_prime],
+                [v * a_prime, v * a_prime, u * a_prime],
+            ];
+            type = LATTICE_TYPE.TRI;
```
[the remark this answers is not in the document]
VsevolodX: Fixed as well

src/lattice/lattice_bravais.js
VsevolodX: Yes. I've noticed that

src/parsers/espresso/espresso.js
```diff
@@ -0,0 +1,389 @@
...
+            break;
+        case 13:
+            type = LATTICE_TYPE.MCLC; // Monoclinic P, base centered
+            break;
+        case -13:
+            type = LATTICE_TYPE.MCLC; // Monoclinic P, unique axis b
+            break;
+        case 14:
+            type = LATTICE_TYPE.TRI;
+            break;
```
timurbazhirov, line 216: I'd isolate to an object and put to settings.js here:

```js
IBRAV_TO_LATTICE_TYPE_MAP = {
     "1": LATTICE_TYPE.CUB,
     ...
     "-13": LATTICE_TYPE.MCLC; // Monoclinic P, unique axis b
}
```

then do `return IBRAV_TO_LATTICE_TYPE_MAP[`${ibrav}`]` or throw error

Twice as little code

VsevolodX: Hm. That's smart. I've adjusted

src/parsers/espresso/espresso.js
```diff
@@ -0,0 +1,389 @@
...
+            [
+                parseFloat(cellParameters[5]),
+                parseFloat(cellParameters[6]),
+                parseFloat(cellParameters[7]),
+            ],
+            [
+                parseFloat(cellParameters[8]),
+                parseFloat(cellParameters[9]),
+                parseFloat(cellParameters[10]),
+            ],
```
timurbazhirov, line 73: All of the above on lines 59-72 can be done via regex group matching

src/utils/parsers/settings.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {String | String[]} values - to be put in the string
+ * @returns {String}
+ */
+export function formatString(str, ...values) {
+    return str.replace(/{(\d+)}/g, (match, index) => {
+        return typeof values[index] !== "undefined" ? values[index] : match;
+    });
+}
+
+const fortranDoubleRegex = "([-+]?\\d*\\.?\\d*(?:[Eed][+-]?\\d+)?)";
```
timurbazhirov, line 15: Here's a good example of how RegEx should be defined and explained https://github.com/Exabyte-io/api-examples/blob/6ca21439ad2239bf82ade75ac0a34ccfd17bfcb0/examples/config.py#L34.
Screen Shot 2023-06-20 at 1 15 14 PM

VsevolodX: Makes sense. Noted.

src/parsers/espresso/espresso.js
VsevolodX: It's a matter of preference. 
I'd do it with named capturing groups like this:
```
    if (match) {
        const vectors = [
            [parseFloat(match.groups.x1), parseFloat(match.groups.y1), parseFloat(match.groups.z1)],
            [parseFloat(match.groups.x2), parseFloat(match.groups.y2), parseFloat(match.groups.z2)],
            [parseFloat(match.groups.x3), parseFloat(match.groups.y3), parseFloat(match.groups.z3)],
        ];
```

While there's example of reverse function which is one liner basically, but personally I find it harder to understand correctly from the first scanning look:
```
const vectorsAsString = _.map(vectors, (v) => {
        return `${s.sprintf("%14.9f", v[0])}\t${s.sprintf("%14.9f", v[1])}\t${s.sprintf(
            "%14.9f",
            v[2],
        )}`;
    }).join("\n");
```

Although, if map functions like this are coding equivalent of Einstein summation, I'll use them

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,247 @@
...
+    if (data.system.ibrav === 0) {
+        cell = getCellConfig(data.cards);
+        cell.type = Lattice.typeFromVectors(cell.vectors); // not implemented yet, defaults to TRI
+    } else {
+        cell = ibravToCellConfig(data.system);
+    }
+
+    const { elements, coordinates, constraints, units } = getAtomicPositions(data.cards);
+    const name = data.control.title;
+    return { cell, elements, coordinates, constraints, units, name };
```
timurbazhirov, line 244: Let's define a class with these properties and have its specific implementation for QE

src/parsers/init.js
```diff
@@ -0,0 +1,24 @@
+class BaseParser {
```
timurbazhirov, line 1: There should StructurParser with `cell`, `elements` etc. as an intermediate data

src/parsers/native_format_parsers.js
```diff
@@ -1,51 +1,100 @@
...
+            return Poscar.fromPoscar(text); // TODO: rewrite when ready to use PoscarParser
+        // return PoscarParser.getIntermediateFormat(text);
+        case NATIVE_FORMATS.QE:
+            return EspressoParser.getIntermediateFormat(text);
+        case NATIVE_FORMATS.UNKNOWN:
             throw new Error(`Unknown format`);
         // TODO:  add more formats
         default:
-            throw new Error(`Unsupported format: ${format}`);
+            throw new Error(`Unsupported format: ${format.name}`);
```
timurbazhirov, line 44: Use https://www.tutorialspoint.com/design_pattern/factory_pattern.htm

src/lattice/lattice_bravais.js
```diff
@@ -122,6 +122,18 @@ export class LatticeBravais {
         return object;
     }
 
+    /**
+     * Return lattice type from vectors array.
+     * @param {Number[][]} array
+     * @returns {String}
+     */
+    // eslint-disable-next-line no-unused-vars
+    static typeFromVectors(array) {
```
timurbazhirov, line 131: Remove

src/parsers/enums.js
```diff
@@ -0,0 +1,14 @@
+export const NATIVE_FORMATS = {
```
timurbazhirov, line 1: STRUCTURAL_INFORMATION_FORMATS

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,292 @@
...
+import { primitiveCell } from "../../../cell/primitive_cell";
+import { Lattice } from "../../../lattice/lattice";
+import math from "../../../math";
+import { FortranParser } from "../../../utils/parsers/fortran";
+import { IBRAV_TO_LATTICE_TYPE_MAP, regex } from "./settings";
+
+export class EspressoParser {
+    constructor(content) {
+        this.content = content;
+        this.intermediateFormat = this.getIntermediateFormat();
```
timurbazhirov, line 12: Remove

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,292 @@
...
+        };
+
+        console.log("intermediateFormat:", intermediateData);
+        return intermediateData;
+    }
+
+    serialize() {
+        const { atomicSpecies } = this.intermediateFormat;
+        const { atomicPositions } = this.intermediateFormat;
+        const { cellParameters } = this.intermediateFormat;
```
timurbazhirov, line 47: Lines 45-47 can be on one line

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,292 @@
...
+        return { atomicSpecies, atomicPositions, cellParameters };
+    }
+}
+
+/**
+ * @summary checks if the given fileContent is in the format of a Quantum ESPRESSO .in file.
+ * @param {String} fileContent
+ * @return {Boolean}
+ */
+function isEspressoFormat(fileContent) {
```
timurbazhirov, line 61: Remove checks for formats

src/parsers/poscar/5/parser.js
```diff
@@ -0,0 +1,128 @@
+import { ATOMIC_COORD_UNITS } from "@exabyte-io/code.js/dist/constants";
```
timurbazhirov, line 1: Remove poscar

src/parsers/init.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    /**
+     * Serializes intermediate format to Object.
+     * @param {Object} intermediateFormat
+     * @returns {{cell: Object, elements: Object, coordinates: Object[], units: String, constraints: Object[], name: String}}
+     */
+    static serialize(intermediateFormat) {
+        const { cell, elements, coordinates, units, constraints, name } = intermediateFormat.data;
+        return { cell, elements, coordinates, units, constraints, name };
```
timurbazhirov, line 19: This should not be inside BaseParser, it's material specific

src/parsers/espresso/7.2/settings.js
```diff
@@ -0,0 +1,61 @@
+import { LATTICE_TYPE } from "../../../lattice/types";
```
timurbazhirov, line 1: This file can stay

timurbazhirov: Remove version association

src/parsers/init.js
```diff
@@ -0,0 +1,28 @@
+class BaseParser {
+    static property = {};
+
+    constructor(content) {
+        this.content = content;
+    }
+
+    static parse() {
```
timurbazhirov, line 8: Should be function of two parameters:

- content
- property_name

https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/src/xmlparser/XMLParser.js

src/parsers/init.js
```diff
@@ -0,0 +1,28 @@
+class BaseParser {
+    static property = {};
+
+    constructor(content) {
+        this.content = content;
```
timurbazhirov, line 5: Should have options only

src/parsers/init.js
```diff
@@ -0,0 +1,28 @@
...
+    static parse() {
+        throw new Error("parse() is not defined");
+    }
+
+    /**
+     * Serializes intermediate format to Object.
+     * @param {Object} intermediateFormat
+     * @returns {{cell: Object, elements: Object, coordinates: Object[], units: String, constraints: Object[], name: String}}
+     */
+    static serialize(intermediateFormat) {
```
timurbazhirov, line 17: Remove

src/parsers/init.js
```diff
@@ -0,0 +1,28 @@
...
+        const { cell, elements, coordinates, units, constraints, name } = intermediateFormat.data;
+        return { cell, elements, coordinates, units, constraints, name };
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    validate() {
```
timurbazhirov, line 23: Remove

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,292 @@
...
+ * @returns {{elements: Object[], coordinates: Object[], constraints: Object[], units: String}}
+ */
+function getAtomicPositions(text) {
+    const atomicSpeciesMatches = Array.from(text.matchAll(regex.atomicSpecies));
+    // eslint-disable-next-line no-unused-vars
+    const atomicSpecies = atomicSpeciesMatches.map((match) => ({
+        element: match[1],
+        mass: parseFloat(match[2]),
+        potential: match[3],
+    }));
```
spectre007, line 129: Can we remove this? It's unused and the `ATOMIC_POSITIONS` block seems to be enough to define elements, coordinates, and constraints.

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,292 @@
...
+        value: [
+            parseFloat(match[2]) * scalingFactor,
+            parseFloat(match[3]) * scalingFactor,
+            parseFloat(match[4]) * scalingFactor,
+        ],
+    }));
+    const constraints = atomicPositionsMatches
+        .filter((match) => match[5] && match[6] && match[7]) // Check if all three constraints exist
+        .map((match, index) => ({
+            id: index,
```
spectre007, line 149: If only a subset of coordinates has constraints, the indices will not be different due to the filtering in line 147. Perhaps we can use `.reduce()` here.

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,292 @@
...
+
+/**
+ * @summary Calculates cell parameters from celldm(i) or A, B, C parameters depending on which are present. Specific to Quantum Espresso.
+ * @param {Number[]} [celldm] - celldm(i) parameters
+ * @param {Number} [a] - A parameter
+ * @param {Number} [b] - B parameter
+ * @param {Number} [c] - C parameter
+ * @returns {Number[]}
+ */
+function getCellConstants(celldm, a, b, c) {
```
spectre007, line 178: Why are `a`, `b`, and `c` function arguments if they are just passed through? I think you could just use `celldm` here and this would also get rid of all the if statements:
```javascript
function getCellConstants(celldm) {
    const a = celldm[1] * coefficients.BOHR_TO_ANGSTROM;
    const b = celldm[2] * celldm[1] * coefficients.BOHR_TO_ANGSTROM;
    const c = celldm[3] * celldm[1] * coefficients.BOHR_TO_ANGSTROM;
    return [a, b, c];
```

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,292 @@
...
+    // Specific case for hexagonal cell in QE docs
+    if (celldm && celldm[4] && celldm[5] && celldm[6]) {
+        alpha = math.acos(celldm[4]);
+        beta = math.acos(celldm[5]);
+        gamma = math.acos(celldm[6]);
+    }
+
+    // Convert radians to degrees which are used in lattice definitions
+    [alpha, beta, gamma] = [alpha, beta, gamma].map((x) =>
+        x === undefined ? x : (x * 180) / math.PI,
```
spectre007, line 216: Is it expected that some of the cell angles are undefined? If so, then I would update the docstring to reflect that`@returns {Array<Number | undefined>}`

src/parsers/espresso/7.2/parser.js
```diff
@@ -0,0 +1,292 @@
...
+ */
+function ibravToCellConfig(system) {
+    const { ibrav, celldm, a, b, c, cosab, cosac, cosbc } = system;
+    if (celldm && a) {
+        throw new Error("Both celldm and A are given");
+    } else if (!celldm && !a) {
+        throw new Error("Missing celldm(1)");
+    }
+
+    if (celldm) celldm.unshift(null); // Bump indices by 1 to make it easier to understand the logic. In QE input file celdm(1) array starts with 1, but parsed starting with 0.
```
spectre007, line 249: I see where you are coming from, but I don't think we should modify the array to use Fortran counting. It might be understandable in this function, but perhaps not so much in subsequent functions which use `celldm`. If you need the correspondence with `celldm(1)`, etc. I would suggest destructuring the array: `const [celldm1, celldm2, celldm3] = celldm;`

src/parsers/espresso/init.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+// eslint-disable-next-line no-unused-vars
+const versionRegexMap = {
+    // TODO: implement version detection
+    5.2: /temp/,
+    "5.*": function () {},
+    "6.*": function () {},
+    "7.*": function () {},
+};
```
spectre007, line 20: I think this can be removed as we are not differentiating versions right now.

src/parsers/espresso/init.js
```diff
@@ -0,0 +1,61 @@
...
+    static getIntermediateFormat(text) {
+        const version = EspressoParser.getVersionByContent(text);
+        const parser = EspressoParser.getParserByVersion(version);
+        return {
+            metadata: {
+                application: APPLICATIONS.ESPRESSO,
+                format: NATIVE_FORMATS.QE,
+                version: EspressoParser.getVersionByContent(text),
+            },
+            data: parser.fromEspressoFormat(text),
```
spectre007, line 56: Following up on the comment from above (https://github.com/Exabyte-io/made.js/pull/81/files#r1251421507), it would make sense to have the parsing functions as class methods of `EspressoParser`.

src/utils/parsers/settings.js
```diff
@@ -0,0 +1,71 @@
...
+    "\\n"; // Ends with a newline character
+
+const stringRegex = "([+\\w.\\-\\/]*)"; // Matches alphanumeric, plus, period, hyphen, and slash characters
+
+const fortranStringRegex =
+    "'" + // Starting single quote
+    "([\\w.\\-\\+\\/ ]*)" + // Matches alphanumeric, period, hyphen, plus, slash, and space characters
+    "'"; // Ending single quote
+
+const fortranArrayRegex =
```
spectre007, line 37: It seems the namelist format supports a few more options for representing arrays. According to [ref1](https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vnc6/index.html#:~:text=You%20can%20alternatively,program%20segment%20above%3A) and [ref2](https://www.cenapad.unicamp.br/parque/manuais/Xlf/lr96.HTM), they could also take the following form:
```
&FOO
A(3) = 2, 10, 15, 16
mat(2, 2) = 2.2
/
```
Although unlikely for Quantum Espresso, we should ensure the array regexes include such cases.

spectre007 (commented):
In addition to my comments a few general things:
- the version directories `7.2` and `5` are now redundant, we can move their content in the parent directory
- `init.js` is not very descriptive. If it defines base parsers then maybe better use `base.js` or sth like that
- similarly `settings.js` mostly contains regexes, so then sth like `regexes.js` would be more descriptive

src/parsers/espresso/7.2/parser.js
VsevolodX: In QE A, B and C parameters can be supplied instead of celldm list. So, it's possible that this function receives only a,b,c and no celldm.

src/parsers/espresso/7.2/parser.js
VsevolodX: Ok

src/parsers/espresso/7.2/parser.js
VsevolodX: Ok, will change. I thought constraints either present on all atoms or not at all

src/parsers/espresso/7.2/parser.js
VsevolodX: I've shifted indices in functions where they are used with a comment added, as we discussed

timurbazhirov:
Labeled to revisit later and closing

PR #82, opened 2023-07-11
Feature/sof 6598 clean

VsevolodX:
Draft PR.


src/parsers/init.js
```diff
@@ -0,0 +1,39 @@
...
+        this.options = options;
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    parse() {
+        throw new Error("parse() is implemented in children");
+    }
+}
+
+export class MaterialParser extends BaseParser {
```
timurbazhirov, line 13: Move to `structure.js`.

You should define the function that forms the material here in this class as something like:

```js
parseMaterial() {

   const cell = this.getCell();
   const elements = this.getElements();
   const coordinates = this.getCoordinates();
   const constraints = this.getConstraints();

   const lattice = Lattice.fromVectors({
            a: cell.cell[0],
            b: cell.cell[1],
            c: cell.cell[2],
            type: cell.type,
        });

        const basis = new ConstrainedBasis({
            elements,
            coordinates,
            units,
            type: cell.type,
            cell: lattice.vectorArrays,
            constraints,
        });

        return {
            lattice: lattice.toJSON(),
            basis: basis.toJSON(),
            name: this.data.control.title,
            isNonPeriodic: false,
        };
```

then in child classes define the corresponding functions and reuse parseMaterial through inheritance without touching

src/parsers/espresso/parser.js
```diff
@@ -0,0 +1,259 @@
...
+            }));
+        return { elements, coordinates, constraints, units: _units };
+    }
+
+    /**
+     * @summary Return units and scaling factor according to Quantum ESPRESSO docs
+     * @param {String} units - units from ATOMIC_POSITIONS card
+     * @returns {{_units: String, scalingFactor: Number}}
+     */
+    getScalingFactor(units) {
```
timurbazhirov, line 233: getCoordinateUnitsScalingFactor

src/parsers/espresso/parser.js
```diff
@@ -0,0 +1,259 @@
...
+import { ConstrainedBasis } from "../../basis/constrained_basis";
+import { primitiveCell } from "../../cell/primitive_cell";
+import { Lattice } from "../../lattice/lattice";
+import math from "../../math";
+import { MaterialParser } from "../init";
+import { FortranParser } from "../utils/fortran";
+import { IBRAV_TO_LATTICE_TYPE_MAP, regex } from "./settings";
+
+export class ESPRESSOMaterialParser extends MaterialParser {
+    parseMaterial(content) {
```
timurbazhirov, line 12: Remove per my comment below

src/parsers/espresso/parser.js
```diff
@@ -0,0 +1,259 @@
...
+
+import { ConstrainedBasis } from "../../basis/constrained_basis";
+import { primitiveCell } from "../../cell/primitive_cell";
+import { Lattice } from "../../lattice/lattice";
+import math from "../../math";
+import { MaterialParser } from "../init";
+import { FortranParser } from "../utils/fortran";
+import { IBRAV_TO_LATTICE_TYPE_MAP, regex } from "./settings";
+
+export class ESPRESSOMaterialParser extends MaterialParser {
```
timurbazhirov, line 11: Reuse Fortran parser capability through a `FotranParserMixin` and multiple inheritance

src/parsers/utils/fortran.js
```diff
@@ -0,0 +1,115 @@
...
+    const match = regex.fortran.cards.exec(text);
+    // eslint-disable-next-line prefer-destructuring
+    output.cards = match[0];
+    return output;
+}
+
+export class FortranParser extends BaseParser {
+    // eslint-disable-next-line class-methods-use-this
+    parse(content) {
+        return parseFortranFile(content);
```
timurbazhirov, line 113: Doesn't make sense - just implement inside the class - move lines 97-107 here

src/parsers/utils/fortran.js
```diff
@@ -0,0 +1,115 @@
...
+        throw new Error("Incorrect fortran file");
+    }
+
+    const match = regex.fortran.cards.exec(text);
+    // eslint-disable-next-line prefer-destructuring
+    output.cards = match[0];
+    return output;
+}
+
+export class FortranParser extends BaseParser {
```
timurbazhirov, line 110: Should be defined as a mixin per the above

VsevolodX: The FortranParser is a mixin, not a class itself?

src/parsers/init.js
VsevolodX: ok

src/parsers/utils/fortran.js
VsevolodX: To test `FortranParserMixin`'s `parseNamelists()` i've created a `TestClass` extending `BaseParser `with `FortranParserMixin`

src/parsers/espresso/parser.js
```diff
@@ -0,0 +1,255 @@
...
+        return this.getCellConfig(this.data.cards);
+    }
+
+    getElements() {
+        const { elements } = this.getAtomicPositions(this.data.cards);
+        return elements;
+    }
+
+    getCoordinates() {
+        const { coordinates } = this.getAtomicPositions(this.data.cards);
```
timurbazhirov, line 27: This doesn't make sense - actually split lines 203-210 here, and similarly for elements, etc.

src/parsers/utils/fortran.js
```diff
@@ -0,0 +1,110 @@
...
+ * @param {Function | NumberConstructor} type - The type of the value.
+ * @param {Boolean} [isArray=false] - Whether to treat the value as an array.
+ *
+ * @returns {Array} The extracted pairs. Each pair is represented as an array,
+ *                  where the first element is the key and the second element is the value.
+ *                  If isArray is true, the value is an array where the first element
+ *                  is the index of the Fortran array element and the second element is the value.
+ * @throws {Error} If an invalid type is provided.
+ */
+function extractPairs(data, regexPattern, type, isArray) {
```
timurbazhirov, line 24: Make all these to be class methods too, inside `FortranParserMixin`

src/parsers/espresso/parser.js
VsevolodX: It did make sense when the idea of intermediate format was still around

src/parsers/espresso/parser.js
```diff
@@ -0,0 +1,255 @@
...
+        } else if (!celldm && !a) {
+            throw new Error("Missing celldm(1)");
+        }
+
+        const type = this.ibravToCellType(ibrav);
+        const [_a, _b, _c] = this.getLatticeConstants(celldm, a, b, c);
+        const [alpha, beta, gamma] = this.getLatticeAngles(celldm, cosbc, cosac, cosab);
+
+        const lattice = new Lattice({
+            a: _a,
```
timurbazhirov, line 109: This thing with `_a`, `_b`, `_c`, `_units` just seems odd - why not use just a, b, c and `let` instead of `const`?

src/parsers/utils/fortran.js
```diff
@@ -0,0 +1,110 @@
+import { regex } from "./settings";
```
timurbazhirov, line 1: This file doesn't need to be in utils, can be one level above

src/parsers/native_format_parsers.js
```diff
@@ -1,26 +1,19 @@
+import { STRUCTURAL_INFORMATION_FORMATS } from "./enums";
```
timurbazhirov, line 1: this file really contains the Factory - should be

```JS
const parser = ParserFactor.getParserByDetectingTextFormat(text);
return parser.parse(text);
```

src/parsers/utils/fortran.js
VsevolodX: Are you talking about `fortran.js` or `settings.js`?

src/parsers/native_format_parsers.js
VsevolodX: If we want to implement this now in here, we need to touch and change calls to JSONParser, PoscarParser etc and create new format detection function

src/parsers/espresso/parser.js
```diff
@@ -0,0 +1,271 @@
...
+
+import { primitiveCell } from "../../cell/primitive_cell";
+import { Lattice } from "../../lattice/lattice";
+import math from "../../math";
+import { MaterialParser } from "../structure";
+import { FortranParserMixin } from "../utils/fortran";
+import { IBRAV_TO_LATTICE_TYPE_MAP, regex } from "./settings";
+
+export class ESPRESSOMaterialParser extends mix(MaterialParser).with(FortranParserMixin) {
+    parse(content) {
```
timurbazhirov, line 12: I think ESLint rules should require a newline after class definition

src/parsers/utils/fortran.js
```diff
@@ -0,0 +1,127 @@
...
+        /**
+         * Parses Fortran namelists and cards data from a string.
+         *
+         * @summary Parses Fortran namelists and cards data from a QE input file string.
+         * @param {String} content - The text to parse.
+         * @throws {Error} If no namelist data is found in `text`.
+         * @throws {Error} If no cards data is found in `text`.
+         * @returns {Object} An object containing the parsed namelist and cards data. The exact structure of this object will depend on the structure of the namelist and cards data in `text`.
+         */
+        parseNamelists(content) {
```
timurbazhirov, line 20: Let's wrap up the functions in `fortran...` to make it easier to find where they came from?

fortranParseNamelist
fortranExtractPairs
etc.

src/parsers/utils/fortran.js
```diff
@@ -0,0 +1,127 @@
...
+         * @param {Function | NumberConstructor} type - The type of the value.
+         * @param {Boolean} [isArray=false] - Whether to treat the value as an array.
+         *
+         * @returns {Array} The extracted pairs. Each pair is represented as an array,
+         *                  where the first element is the key and the second element is the value.
+         *                  If isArray is true, the value is an array where the first element
+         *                  is the index of the Fortran array element and the second element is the value.
+         * @throws {Error} If an invalid type is provided.
+         */
+        extractPairs(data, regexPattern, type, isArray) {
```
timurbazhirov, line 49: The naming is hard to understand and there are no examples in the docstring and/or tests - wth is a "pair"?? Why is there a "pair" and a "key-value pair" that are separate?? Let's add examples

src/parsers/utils/fortran.js
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        /**
+         * @summary Extracts an array of the key value pairs from a Fortran namelist.
+         * @param {String} data
+         * @returns {Object[]}
+         */
+        extractKeyValuePairs(data) {
+            const output = {};
+            const numberPairs = this.extractPairs(data, regex.fortran.numberKeyValue, Number); // TODO: Fix to convert numbers like 1.234D-567
```
timurbazhirov, line 70: Let's not use TODO comments on the same line - put above

src/parsers/espresso/parser.js
```diff
@@ -0,0 +1,271 @@
...
+     * @returns {{id: Number, value: Number[]}[]}
+     */
+    getCoordinates() {
+        const text = this.data.cards;
+        const atomicPositionsMatches = Array.from(text.matchAll(regex.atomicPositions));
+        const units = text.match(regex.atomicPositionsUnits)[1];
+        const { scalingFactor } = this.getCoordinatesUnitsScalingFactor(units);
+        const coordinates = atomicPositionsMatches.map((match, index) => ({
+            id: index,
+            value: [
```
timurbazhirov, line 50:
```JS
value: match.slice(2, 5).map(value => parseFloat(value) * scalingFactor)
```

src/parsers/espresso/parser.js
```diff
@@ -0,0 +1,271 @@
...
+    /**
+     * @summary Return units and scaling factor according to Quantum ESPRESSO 7.2 docs
+     * @param {String} units - units from ATOMIC_POSITIONS card
+     * @returns {{units: String, scalingFactor: Number}}
+     */
+    getCoordinatesUnitsScalingFactor(units) {
+        let _units, scalingFactor;
+        switch (units) {
+            case "alat":
+                scalingFactor = 1.0;
```
timurbazhirov, line 249: Just set it at the top on line 246

src/parsers/espresso/parser.js
VsevolodX: ok

src/parsers/utils/fortran.js
VsevolodX: I think it was done in order to shorten the function length and isolate repeated code

src/parsers/espresso/parser.js
VsevolodX: Not in my setup, which is pulled from /dev branch

src/parsers/utils/fortran.js
VsevolodX: Changed to KeyValuePair(s)

PR #83, opened 2023-08-25
SOF-6588 feat: add getConsistencyChecks

VsevolodX:
This is the draft approach to test the waters. Checks for atom feasibility (verification? validation?) should happen in Made.Material. The generation of check result message or warning should be done in a separate class.

### DIFF
--- a/src/basis/basis.js
+++ b/src/basis/basis.js
@@ -453,8 +453,8 @@ export class Basis {
         coordinates.forEach((entry1) => {
             coordinates.forEach((entry2) => {
                 if (entry1.id === entry2.id) return;
-                const el1 = elements.find((el) => el.id === entry1.id).value;
-                const el2 = elements.find((el) => el.id === entry2.id).value;
+                const el1 = elements.getElementByIndex(entry1.id);
+                const el2 = elements.getElementByIndex(entry2.id);
                 // temporary value for overlap approximation, where atoms most certainly can't be located
                 const overlapCoefficient = 0.75;
                 const tolerance =
@@ -466,8 +466,8 @@ export class Basis {
                     overlaps.push({
                         id1: entry1.id,
                         id2: entry2.id,
-                        element1: elements.find((el) => el.id === entry1.id).value,
-                        element2: elements.find((el) => el.id === entry2.id).value,
+                        element1: el1,
+                        element2: el2,
                     });
                 }
             });
--- a/src/basis/basis.js
+++ b/src/basis/basis.js
@@ -455,8 +455,8 @@ export class Basis {
         coordinates.forEach((entry1, i) => {
             for (let j = i + 1; j < coordinates.length; j++) {
                 const entry2 = coordinates[j];
-                const el1 = elements[i];
-                const el2 = elements[j];
+                const el1 = elements[i].value;
+                const el2 = elements[j].value;
 
                 const tolerance =
                     overlapCoefficient *
@@ -475,7 +475,6 @@ export class Basis {
         });
 
         this.toCrystal();
-
         return overlaps;
     }
 
--- a/src/basis/basis.js
+++ b/src/basis/basis.js
@@ -447,14 +447,14 @@ export class Basis {
     getOverlappingAtoms() {
         // to simplify calculations, convert to cartesian coordinates
         this.toCartesian();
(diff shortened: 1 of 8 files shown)


src/material.js
```diff
@@ -319,4 +319,31 @@ export class Material extends HasMetadataNamedDefaultableInMemoryEntity {
...
+        overlappingGroups.forEach((group) => {
+            group.forEach((atom) => {
+                const otherAtoms = group
+                    .filter((other) => other.id !== atom.id)
+                    .map((other) => `${other.element} (Line ${other.id + 1})`)
+                    .join(", ");
+
+                checks.push({
+                    type: "warning",
+                    message: `${atom.element} (Line ${atom.id + 1}) overlaps with ${otherAtoms}`,
```
VsevolodX, line 341: I'm thinking turning this into more generic message from an enum, to use like `messages.atomsOverlap` and define it all in a dedicated property.

src/basis/basis.js
```diff
@@ -440,6 +440,42 @@ export class Basis {
...
+     */
+    getOverlappingAtoms() {
+        // to simplify calculations, convert to cartesian coordinates
+        const _basis = this;
+        _basis.toCartesian();
+        const { coordinates, elements } = _basis;
+        const overlaps = [];
+
+        coordinates.forEach((entry1) => {
+            coordinates.forEach((entry2) => {
```
timurbazhirov, line 455: We only need to loop over the remaining atoms, not all of them again

src/basis/basis.js
```diff
@@ -440,6 +440,42 @@ export class Basis {
...
+                const tolerance =
+                    overlapCoefficient *
+                    (getElementAtomicRadius(el1) + getElementAtomicRadius(el2)); // in angstroms
+
+                const distance = math.vDist(entry1.value, entry2.value);
+                if (distance < tolerance) {
+                    overlaps.push({
+                        id1: entry1.id,
+                        id2: entry2.id,
+                        element1: elements.find((el) => el.id === entry1.id).value,
```
timurbazhirov, line 469: use el1 and el2

where is the schema for overlaps defined?

src/basis/basis.js
```diff
@@ -440,6 +440,42 @@ export class Basis {
...
+        const _basis = this;
+        _basis.toCartesian();
+        const { coordinates, elements } = _basis;
+        const overlaps = [];
+
+        coordinates.forEach((entry1) => {
+            coordinates.forEach((entry2) => {
+                if (entry1.id === entry2.id) return; // Don't compare an atom with itself
+                const el1 = elements.find((el) => el.id === entry1.id).value;
+                const el2 = elements.find((el) => el.id === entry2.id).value;
```
timurbazhirov, line 458: Maybe use indes in forEach and  `getElementByIndex`?

src/material.js
```diff
@@ -319,4 +319,44 @@ export class Material extends HasMetadataNamedDefaultableInMemoryEntity {
...
+
+        const basis = this.Basis;
+        const overlappingAtomsGroups = basis.getOverlappingAtoms();
+
+        overlappingAtomsGroups.forEach(({ id1, id2, element1, element2 }) => {
+            const key = `basis.coordinates.${id1}`;
+            keys.add(key);
+            messages.push({
+                key,
+                checkName: "atomsOverlap",
```
timurbazhirov, line 354: Why do we hardcode "atomsOverlap"?

src/basis/basis.js
```diff
@@ -440,6 +440,42 @@ export class Basis {
         return math.precise(latticeSize, 4);
     }
 
+    /**
+     * @summary function returns an array of overlapping atoms within specified tolerance.
+     * @returns {{element: String, id: Number}[]}
+     */
+    getOverlappingAtoms() {
```
timurbazhirov, line 447: We should make this configurable and turn off when the number of atoms exceeds a certain value - i.e. 1000

src/material.js
VsevolodX: Ideally, it should be something like `checkNamesEnum.material.basis.atomsOverlap` stored in some centralized repository. This sounds like definitions, core to the entities that get checked -- code.js ?
But for now not to overoptimize I left it like that.

PR #93, opened 2023-11-08
SOF-7097: Material mixin

k0stik:


timurbazhirov (approved):

src/material.ts
```diff
@@ -62,316 +63,341 @@ export const defaultMaterialConfig = {
...
+            return this.prop<FileSourceSchema>("src");
+        }
 
-    updateFormula() {
-        this.setProp("formula", this.Basis.formula);
-        this.setProp("unitCellFormula", this.Basis.unitCellFormula);
-    }
+        set src(src: FileSourceSchema) {
+            this.setProp("src", src);
+        }
```
seankwarren, line 121: These dont seem to have existed in previous Material class. Why are they added in this task?

timurbazhirov: ?

k0stik: There was 
```
abstract src: {
        extension: string;
        text: string;
    };
```
But this was not 100% correct, because we are using `src` here in `getAsPOSCAR`, though we didn't define how to fetch it

PR #108, opened 2024-03-04
fix: ts tests

k0stik:


src/cell/cell.ts
```diff
@@ -1,7 +1,8 @@
+import { math } from "@exabyte-io/code.js/dist/math";
+
 import { Coordinate } from "../basis/types";
 import constants from "../constants";
 import { Vector, VectorsAsArray } from "../lattice/types";
-import math from "../math";
```
timurbazhirov, line 4: Did we update other usage of math in made too?

PR #112, opened 2024-03-21
chore: make the package dual-use with both PY and Typescript/JS runtimes

timurbazhirov:


VsevolodX (approved):
Checked out locally. Package does install and work in MD

PR #122, opened 2024-04-16
feat: basic tool functions

timurbazhirov:


src/py/mat3ra/made/tools/convert.py
```diff
@@ -0,0 +1,174 @@
...
+        ase_atoms (Atoms): An ASE Atoms object.
+
+    Returns:
+        dict: A dictionary containing the material information in ESSE format.
+    """
+    structure = AseAtomsAdaptor.get_structure(ase_atoms)
+    return from_pymatgen(structure)
+
+
+def convert_material_args_kwargs_to_atoms(func: Callable) -> Callable:
```
VsevolodX, line 158: I like this approach to generalize functions to take arguments in any format

tests/py/unit/test_tools_convert.py
```diff
@@ -0,0 +1,71 @@
...
+from ase import Atoms
+from ase.build import bulk
+from mat3ra.made.material import Material
+from mat3ra.made.tools.convert import from_ase, from_poscar, from_pymatgen, to_ase, to_poscar, to_pymatgen
+from pymatgen.core.structure import Element, Lattice, Structure
+
+PYMATGEN_LATTICE = Lattice.from_parameters(a=3.84, b=3.84, c=3.84, alpha=120, beta=90, gamma=60)
+PYMATGEN_STRUCTURE = Structure(PYMATGEN_LATTICE, ["Si", "Si"], [[0, 0, 0], [0.75, 0.5, 0.75]])
+
+POSCAR_CONTENT = """Si2
```
VsevolodX, line 11: As I understand, this is temporary, and in the future assets and fixtures will be moved to a separate location in standardized manner?

VsevolodX (approved):
LGTM as a base repo

src/py/mat3ra/made/tools/convert.py
timurbazhirov: Yes, the idea is that all tools should be applicable to Material class instances and the conversion should be done in the decorator layer, if needed.

tests/py/unit/test_tools_convert.py
timurbazhirov: Yes

src/py/mat3ra/made/tools/convert.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    Converts material object in ESSE format to an ASE Atoms object.
+
+    Args:
+        material_or_material_data (dict|class): A dictionary containing the material information in ESSE format.
+
+    Returns:
+        Any: An ASE Atoms object.
+    """
+    # TODO: check that atomic labels are properly handled
```
timurbazhirov, line 150: @VsevolodX FYSE - we will need to check that this is working b/c of the difference in Py3.8 toolchain: https://github.com/Exabyte-io/made/pull/122/files#diff-92d73dbc1fad849082fdef64b09c5b6e8d28bf7d2324a1bd220c4d3b32c2c2e1R48. If problematic, we will need to use native ASE conversion (like in to_pymatgen) instead of using the pymatgen adapter.

VsevolodX: Ok

PR #131, opened 2024-05-16
feature/SOF-7372 update: Generalize strain matching algorithm usage

VsevolodX:
- update: rename zsl -> strain matching
- chore: linter fix
- update: add enum
- update: return builder based on name + test

### DIFF
--- a/src/py/mat3ra/made/tools/build/interface.py
+++ b/src/py/mat3ra/made/tools/build/interface.py
@@ -55,7 +55,7 @@ class StrainMatchingParameters:
 class InterfaceBuilderSettings:
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


src/py/mat3ra/made/tools/build/interface.py
```diff
@@ -9,6 +9,20 @@
 from pymatgen.analysis.interfaces.coherent_interfaces import CoherentInterfaceBuilder, ZSLGenerator, Interface
 from ..convert import convert_atoms_or_structure_to_material, decorator_convert_material_args_kwargs_to_structure
 
+TerminationType = Tuple[str, str]
```
VsevolodX, line 12: Moved types and enums definitions on top

PR #140, opened 2024-06-28
Update/SOF-7381 Import and reexport all 3rd party from one place

VsevolodX:
ASE and Pymatgen are handled.


src/py/mat3ra/made/tools/third_party.py
```diff
@@ -0,0 +1,41 @@
+from ase import Atoms as ASEAtoms
+from ase.build.supercells import make_supercell as ase_make_supercell
+from ase.calculators.calculator import Calculator as ASECalculator
+from ase.calculators.emt import EMT
```
timurbazhirov, line 4: ASECalculatorEMT

timurbazhirov (approved):

PR #142, opened 2024-07-03
Feature/SOF-7380 feature: add vacuum adjustment functions

VsevolodX:


src/py/mat3ra/made/tools/modify.py
```diff
@@ -323,3 +319,42 @@ def condition(coordinate):
...
+
+    Args:
+        material (Material): The material object to set the vacuum thickness.
+        vacuum (float): The total thickness of the vacuum in angstroms to set above the material.
+
+    Returns:
+        Material: The material object with the vacuum thickness set.
+    """
+    atoms = to_ase(translate_to_bottom(material))
+    new_c = max(atoms.positions[:, 2]) + vacuum
```
timurbazhirov, line 354: With zero vacuum this will lead to the first and last layer being at the same coordinate, it seems no?

src/py/mat3ra/made/tools/modify.py
```diff
@@ -323,3 +319,42 @@ def condition(coordinate):
     return filter_by_condition_on_coordinates(
         material, condition, use_cartesian_coordinates=use_cartesian_coordinates, invert_selection=invert_selection
     )
+
+
+def add_vacuum(material: Material, vacuum: float = 5.0) -> Material:
```
timurbazhirov, line 324: The selection of functions is a bit weird, usually we would follow a CRUD, get-set, or add-remove pattern. The `add` and `set` pattern does not fit either.

tests/py/unit/fixtures.py
```diff
@@ -220,3 +220,50 @@
     },
     "isUpdated": True,
 }
+
+SI_SLAB_VACUUM = {
```
timurbazhirov, line 224: Could we reuse SI_SLAB and adjust certain values here to avoid copy-paste?

src/py/mat3ra/made/tools/modify.py
VsevolodX: Yes, this is somewhat dangerous operation. You get what you ask for. Because there's no general way to set a distance correctly: covalent radius of last atoms? average interlayer distance in the material in Z? Distance between A and B materials in interface? None will fit all the possibilities

src/py/mat3ra/made/tools/modify.py
VsevolodX: For software. For material it makes sense to "add vacuum", "set vacuum" implies that there's no vacuum by default and we set the exact value. Maybe "remove vacuum" is more fair.

src/py/mat3ra/made/tools/modify.py
VsevolodX: Ok. The covalent radius of the topmost atom sounds like a good middle ground. We need to import periodic_table into py for that

src/py/mat3ra/made/tools/modify.py
```diff
@@ -35,22 +34,33 @@ def filter_by_label(material: Material, label: Union[int, str]) -> Material:
     return new_material
 
 
-@decorator_convert_material_args_kwargs_to_structure
-def translate_to_bottom(structure: PymatgenStructure, use_conventional_cell: bool = True):
+def translate_atoms(
```
timurbazhirov, line 37: let's have two functions maybe - `translate_by_vector` and `translate_to_z_level` to have separate - too much logic in the if below.

z_level = bottom by default

src/py/mat3ra/made/tools/modify.py
```diff
@@ -35,22 +34,33 @@ def filter_by_label(material: Material, label: Union[int, str]) -> Material:
...
     Returns:
-        Structure: The normalized pymatgen Structure object.
+        Material: The translated material object.
     """
-    if use_conventional_cell:
-        structure = PymatgenSpacegroupAnalyzer(structure).get_conventional_standard_structure()
-    structure = translate_to_bottom_pymatgen_structure(structure)
-    return structure
+    atoms = to_ase(material)
+    min_z = min(atoms.positions[:, 2])
```
timurbazhirov, line 53: We can add `get_max_z` and `get_min_z` to analyze also and act on atoms directly

src/py/mat3ra/made/tools/modify.py
```diff
@@ -323,3 +333,55 @@ def condition(coordinate):
     return filter_by_condition_on_coordinates(
         material, condition, use_cartesian_coordinates=use_cartesian_coordinates, invert_selection=invert_selection
     )
+
+
+def add_vacuum(material: Material, vacuum: float = 5.0, top=True, bottom=False) -> Material:
```
timurbazhirov, line 338: maybe add_on_top, add_to_bottom

src/py/mat3ra/made/tools/modify.py
VsevolodX: I'm reusing existing functionality

timurbazhirov: ??

src/py/mat3ra/made/tools/modify.py
```diff
@@ -35,22 +34,46 @@ def filter_by_label(material: Material, label: Union[int, str]) -> Material:
...
     Returns:
-        Structure: The normalized pymatgen Structure object.
+        Material: The translated material object.
     """
-    if use_conventional_cell:
-        structure = PymatgenSpacegroupAnalyzer(structure).get_conventional_standard_structure()
-    structure = translate_to_bottom_pymatgen_structure(structure)
-    return structure
+    atoms = to_ase(material)
+    min_z = min(atoms.positions[:, 2])
```
timurbazhirov, line 50: Expressions like `[:,2]` are hard to read, I suggest we define helpers

```python
min_z = (from ..analyze ) get_atomic_coordinates_min_z(material)
max_z = ...
then instead of external atoms.translate use `traslate_by_vector([0, 0, 1-max_z]) etc.

This way we isolate the use of atoms to a single place
```

src/py/mat3ra/made/tools/modify.py
```diff
@@ -323,3 +346,55 @@ def condition(coordinate):
...
+    Args:
+        material (Material): The material object to set the vacuum thickness.
+        from_top (bool): Whether to remove vacuum from the top.
+        from_bottom (bool): Whether to remove vacuum from the bottom.
+        fixed_padding (float): The fixed padding of vacuum to add to avoid collisions in pbc (in angstroms).
+
+    Returns:
+        Material: The material object with the vacuum thickness set.
+    """
+    atoms = to_ase(translate_to_z_level(material, z_level="bottom"))
```
timurbazhirov, line 390: We should not need ASE in this function anymore

src/py/mat3ra/made/tools/analyze.py
```diff
@@ -229,3 +229,47 @@ def get_atom_indices_with_condition_on_coordinates(
...
+    new_material = material.clone()
+    if use_cartesian_coordinates:
+        new_basis = new_material.basis
+        new_basis.to_cartesian()
+        new_material.basis = new_basis
+    coordinates = new_material.basis.coordinates.to_array_of_values_with_ids()
+    return min([coord.value[2] for coord in coordinates])
+
+
+def get_atomic_coordinates_max_z(
```
timurbazhirov, line 256: Code duplicate - maybe let's have `get_atomic_coordinates_extremum(extremum_type="max", axis="z")`

src/py/mat3ra/made/tools/modify.py
timurbazhirov: Still needs to be done

VsevolodX: Ok

timurbazhirov (approved):

PR #155, opened 2024-08-24
feature/SOF 7442 feat: get top surface atoms

VsevolodX:
- feat: add kd tree surface atoms calculation
- update: add z factor to look deeper
- update: small adjustment
- update :add missing handy function
- update: speedup using np
- fix: get more of the island by selecting deeper
- fix: adjust test since the island creation fixed
- feat: add passivate top function

### DIFF
--- a/src/py/mat3ra/made/tools/analyze.py
+++ b/src/py/mat3ra/made/tools/analyze.py
@@ -340,13 +340,14 @@ def get_atomic_coordinates_extremum(
     return getattr(np, extremum)(values)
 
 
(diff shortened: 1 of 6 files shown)


src/py/mat3ra/made/tools/analyze.py
```diff
@@ -339,6 +340,35 @@ def get_atomic_coordinates_extremum(
     return getattr(np, extremum)(values)
 
 
+def get_surface_atoms_indices(material: Material, distance_threshold: float = 2.5, depth: float = 5) -> List[int]:
```
VsevolodX, line 343: For now only top surface considered, looking perpendicularly to AB plane.

PR #156, opened 2024-08-27
feature/SOF-7425 feat: add displacement for film/substrate

VsevolodX:


src/py/mat3ra/made/tools/modify.py
```diff
@@ -462,3 +464,39 @@ def rotate_material(material: Material, axis: List[int], angle: float) -> Materi
     atoms.wrap()
 
     return Material(from_ase(atoms))
+
+
+def displace_interface(
```
timurbazhirov, line 469: displace_interface_component(interface, displacement, component=InterfacePartsEnum.FILM, use_cartesian_coordinates=True)

Should use displacement in angstroms by default

tests/py/unit/fixtures.py
```diff
@@ -61,6 +61,180 @@
 INTERFACE_STRUCTURE.interface_properties = INTERFACE_PROPERTIES_MOCK
 INTERFACE_NAME = "Cu4(001)-Si8(001), Interface, Strain 0.062pct"
 
+GRAPHENE_NICKEL_INTERFACE = {
```
timurbazhirov, line 64: Let's create an interface through a function here and only check what we need

src/py/mat3ra/made/tools/modify.py
```diff
@@ -462,3 +464,39 @@ def rotate_material(material: Material, axis: List[int], angle: float) -> Materi
...
+        interface (Material): The interface Material object.
+        displacement (List[float]): The displacement vector in angstroms or crystal coordinates.
+        label (InterfacePartsEnum): The label of the atoms to displace ("substrate" or "film").
+        use_cartesian_coordinates (bool): Whether to use cartesian coordinates.
+
+    Returns:
+        Material: The displaced material object.
+    """
+    new_material = interface.clone()
+    if use_cartesian_coordinates:
```
timurbazhirov, line 488: We can also check whether it already is in cartesian or not to avoid unnecessary call to `to_cartesian`

VsevolodX: actually, the material method needs this check -- we already have this check for basis.to_cartesian() and basis.to_crystal()

tests/py/unit/fixtures.py
VsevolodX: Ok. We decided to revert back to JSON due to erroneous interface creation during pytest

timurbazhirov (approved):

PR #172, opened 2024-11-05
update/SOF 7414

VsevolodX:
- update: add switch with factory based on config


src/py/mat3ra/made/tools/build/defect/builders.py
```diff
@@ -44,6 +44,7 @@
     PointDefectPairConfiguration,
 )
 from .factories import DefectBuilderFactory
+from pymatgen.analysis.adsorption import AdsorbateSiteFinder
```
timurbazhirov, line 47: third_party

VsevolodX: we decided to move those sites functionalities into other task

timurbazhirov (approved):

PR #183, opened 2024-12-26
feature/SOF 7530

VsevolodX:
- update: fix simple interface
- chore: docsting

### DIFF
--- a/src/py/mat3ra/made/tools/build/interface/builders.py
+++ b/src/py/mat3ra/made/tools/build/interface/builders.py
@@ -63,7 +63,7 @@ class InterfaceBuilder(BaseBuilder):
 ########################################################################################
 class SimpleInterfaceBuilderParameters(InterfaceBuilderParameters):
     scale_film: bool = True  # Whether to scale the film to match the substrate
-    slab_as_is: bool = False  # Whether to use the substrate as is as a slab instead of creating a slab
+    create_slabs: bool = False  # Whether to use the substrate as is as a slab instead of creating a slab
 
 
 class SimpleInterfaceBuilder(ConvertGeneratedItemsASEAtomsMixin, InterfaceBuilder):
@@ -76,9 +76,9 @@ class SimpleInterfaceBuilder(ConvertGeneratedItemsASEAtomsMixin, InterfaceBuilde
     _GeneratedItemType: type(ASEAtoms) = ASEAtoms  # type: ignore
 
     def __preprocess_slab_configuration(
-        self, configuration: SlabConfiguration, termination: Termination, slab_as_is=False
+        self, configuration: SlabConfiguration, termination: Termination, create_slabs=False
     ) -> ASEAtoms:
-        slab = create_slab(configuration, termination) if not slab_as_is else configuration.bulk
+        slab = create_slab(configuration, termination) if create_slabs else configuration.bulk
         ase_slab = to_ase(slab)
 
         niggli_reduce(ase_slab)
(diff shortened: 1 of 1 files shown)


src/py/mat3ra/made/tools/build/interface/builders.py
```diff
@@ -74,15 +75,20 @@ class SimpleInterfaceBuilder(ConvertGeneratedItemsASEAtomsMixin, InterfaceBuilde
...
+    ) -> ASEAtoms:
+        slab = create_slab(configuration, termination) if not slab_as_is else configuration.bulk
         ase_slab = to_ase(slab)
+
         niggli_reduce(ase_slab)
         return ase_slab
 
     @staticmethod
     def __combine_two_slabs_ase(substrate_slab_ase: ASEAtoms, film_slab_ase: ASEAtoms, distance_z: float) -> ASEAtoms:
+        total_c_length = substrate_slab_ase.cell[2][2] + film_slab_ase.cell[2][2] + distance_z
```
timurbazhirov, line 89: total_z_height

src/py/mat3ra/made/tools/build/interface/builders.py
```diff
@@ -102,11 +108,16 @@ def _generate(self, configuration: InterfaceBuilder._ConfigurationType) -> List[
...
-            configuration.substrate_configuration, configuration.substrate_termination
+            configuration.substrate_configuration,
+            configuration.substrate_termination,
+            slab_as_is=self.build_parameters.slab_as_is,
         )
 
         if self.build_parameters.scale_film:
-            film_slab_ase.set_cell(substrate_slab_ase.cell, scale_atoms=True)
+            temp_cell = [substrate_slab_ase.cell[0], substrate_slab_ase.cell[1], film_slab_ase.cell[2]]
+            # Add z of previous height
```
timurbazhirov, line 118: Comment should say: "We scale the film in XY direction only, and use two scaling steps and a temporary cell"

src/py/mat3ra/made/tools/build/interface/builders.py
```diff
@@ -62,7 +62,8 @@ def _update_material_name(self, material: Material, configuration: InterfaceConf
 #                           Simple Interface Builder                                   #
 ########################################################################################
 class SimpleInterfaceBuilderParameters(InterfaceBuilderParameters):
-    scale_film: bool = True
+    scale_film: bool = True  # Whether to scale the film to match the substrate
+    slab_as_is: bool = False  # Whether to use the substrate as is as a slab instead of creating a slab
```
timurbazhirov, line 66: Let's use `create_slabs`, and have it as True by default

src/py/mat3ra/made/tools/build/interface/builders.py
```diff
@@ -62,7 +62,8 @@ def _update_material_name(self, material: Material, configuration: InterfaceConf
 #                           Simple Interface Builder                                   #
 ########################################################################################
 class SimpleInterfaceBuilderParameters(InterfaceBuilderParameters):
-    scale_film: bool = True
+    scale_film: bool = True  # Whether to scale the film to match the substrate
```
timurbazhirov, line 65: Let's put comments above the line where var is defined

timurbazhirov (approved):

src/py/mat3ra/made/tools/build/interface/builders.py
VsevolodX: Here I use this only for substrate. But we can adjust to be for both, but then we would have to create a slab before passing it

src/py/mat3ra/made/tools/build/interface/builders.py
VsevolodX: It's not how we do elsewhere, I think

PR #189, opened 2025-03-26
Feature/SOF-7570

VsevolodX:
Using newly created Pydantic Model classes from ESSE for builder params and configurations


pyproject.toml
```diff
@@ -4,9 +4,9 @@ dynamic = ["version"]
 description = "MAterials DEfinitions and/or MAterials DEsign library."
 readme = "README.md"
 requires-python = ">=3.8"
-license = {file = "LICENSE.md"}
+license = { file = "LICENSE.md" }
```
timurbazhirov, line 7: Unnecessary changes

timurbazhirov (approved):

pyproject.toml
VsevolodX: To be fair -- it's the correct formatting

PR #194, opened 2025-04-08
Feature/sof-7597 - fix tests + rounding

VsevolodX:


dist/js/tools/supercell.js
```diff
@@ -13,6 +13,8 @@ const ADD = math_1.default.add;
 function generateNewBasisWithinSupercell(basis, cell, supercell, supercellMatrix) {
     const oldBasis = basis.clone();
     const newBasis = basis.clone({ isEmpty: true });
+    // oldBasis.cell = cell.vectorsAsArray;
+    // newBasis.cell = supercell.vectorsAsArray;
```
timurbazhirov, line 17: Remove

src/js/basis/basis.ts
```diff
@@ -140,7 +141,7 @@ export class Basis {
     clone(extraContext?: Partial<BasisProps>): Basis {
         return new (this.constructor as typeof Basis)({
             ...this.toJSON(),
-            cell: this.cell,
+            cell: this.cellRounded as ArrayOf3NumberElementsSchema[],
```
timurbazhirov, line 144: We probably should not round here, but in toJSON instead, to stay consistent

src/js/lattice/lattice_bravais.ts
```diff
@@ -103,7 +102,9 @@ export class LatticeBravais implements LatticeImplicitSchema {
                 length: units,
                 angle: "degree",
             },
-        });
+        };
+        // @ts-ignore
```
timurbazhirov, line 106: Shouldn't be needed

tests/js/basis/basis.js
```diff
@@ -29,7 +29,7 @@ describe("Basis", () => {
     });
 
     it("should return jsonified basis", () => {
-        const basis = new Basis(Na4Cl4.basis);
+        const basis = new Material(Na4Cl4).Basis;
```
timurbazhirov, line 32: We need to remove that si-fake above

timurbazhirov (approved):

PR #197, opened 2025-04-25
feature/SOF 7606

VsevolodX:
- **update: fix creation if ids differ**

### DIFF
--- a/src/js/basis/basis.ts
+++ b/src/js/basis/basis.ts
@@ -212,8 +212,18 @@ export class Basis extends InMemoryEntity implements BasisSchema {
         this.units = ATOMIC_COORD_UNITS.crystal as BasisSchema["units"];
     }
 
-    getElementByIndex(idx: number): string {
-        return this._elements.getElementValueByIndex(idx) as string;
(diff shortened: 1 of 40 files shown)


src/js/basis/basis.ts
```diff
@@ -212,8 +212,18 @@ export class Basis extends InMemoryEntity implements BasisSchema {
         this.units = ATOMIC_COORD_UNITS.crystal as BasisSchema["units"];
     }
 
-    getElementByIndex(idx: number): string {
-        return this._elements.getElementValueByIndex(idx) as string;
+    getElementByIndex(idx: number): AtomicElementValue {
+        return this._elements.getElementValueByIndex(idx) as AtomicElementValue;
+    }
+
+    // TODO: should use method from ArrayWithIds
```
timurbazhirov, line 219: Let's use it - why not?

timurbazhirov (approved):

src/js/basis/basis.ts
VsevolodX: Wanted to see if this actually was the problem first.

PR #198, opened 2025-04-28
SOF-7602: mixin cleanups

k0stik:


src/js/types.ts
```diff
@@ -1,20 +1,15 @@
...
-    Lattice: Lattice;
-}
+export type MaterialInterface = MaterialMixin & NamedEntity;
 
 /**
  * @deprecated Import from './types' directory instead
  */
 
-export * from './types/material';
+export * from "./types/material";
```
VsevolodX, line 15: Hm. I think you need to pull latest dev. I moved this export to types/index.ts to collect and export all types from the single file. 
Do you think it's better to do everything in a single types.ts file?

timurbazhirov (approved):

PR #202, opened 2025-05-13
feature/SOF-7598 update: use slab params

VsevolodX:


src/py/mat3ra/made/tools/build/slab/builders.py
```diff
@@ -14,16 +19,13 @@
 from ...third_party import PymatgenSlab, PymatgenSlabGenerator, label_pymatgen_slab_termination
 
 
-class SlabSelectorParameters(BaseModel):
+class SlabSelectorParameters(SlabSelectorParametersSchema):
     termination: Termination
```
timurbazhirov, line 23: Should we have a type for Termination?

timurbazhirov (approved):

PR #211, opened 2025-06-16
feature/SOF-7641 Metadata class

VsevolodX:


src/py/mat3ra/made/tools/build/metadata.py
```diff
@@ -0,0 +1,39 @@
...
+from pydantic import Field
+
+from mat3ra.code.entity import InMemoryEntityPydantic
+
+
+class BaseMetadata(InMemoryEntityPydantic):
+    model_config = {"arbitrary_types_allowed": True, "extra": "allow"}
+
+    @staticmethod
+    def _to_dict(obj: Any) -> dict:
```
timurbazhirov, line 13: Doesn't make sense as a static method here - just create a util

src/py/mat3ra/made/tools/build/slab/helpers.py
```diff
@@ -96,9 +97,10 @@ def create_slab(
 
 def create_slab_if_not(material: Material, default_slab_configuration: SlabConfiguration) -> Material:
     slab = material
-    if not slab.metadata or slab.metadata["build"]["configuration"]["type"] != SlabConfiguration.__name__:
+    metadata = MaterialMetadata(**slab.metadata or {})
```
timurbazhirov, line 100: We can have a property inside material class for metadata

timurbazhirov (approved):

src/py/mat3ra/made/tools/build/slab/helpers.py
VsevolodX: decided not to proceed with this yet

PR #247, opened 2025-08-20
update: make sure to get most optimal interfaces

VsevolodX:


src/py/mat3ra/made/tools/analyze/interface/zsl.py
```diff
@@ -84,38 +84,49 @@ def zsl_match_holders(self) -> List[ZSLMatchHolder]:
...
             match_holders.append(match_holder)
 
+        match_holders = self.purge_equivalent_matches(match_holders)
+
         # sort matches by strain in ascending order, then for each equal strain by area in ascending order
         match_holders = self.sort_by_strain_then_area(match_holders)
 
         return match_holders
 
+    def purge_equivalent_matches(self, match_holders: List[ZSLMatchHolder]) -> List[ZSLMatchHolder]:
```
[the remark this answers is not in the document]
VsevolodX: It's a private function, matches themselves are unique, because really we don't need equivalent ones.

tests/py/unit/test_tools_analyze.py
```diff
@@ -97,13 +98,43 @@ def test_lattice_material_analyzer(
...
     assert_two_entities_deep_almost_equal(primitive_cell_generated, expected_primitive_material_config)
 
 
+@pytest.mark.parametrize(
+    "material_config, is_flipped",
+    [
+        (BULK_Si_CONVENTIONAL, False),
+    ],
+)
+def test_basis_analyzer_(material_config, is_flipped):
```
timurbazhirov, line 119: Each analyzer should have its own test file and we need to test each function for the analyzer here

src/py/mat3ra/made/tools/analyze/basis.py
```diff
@@ -0,0 +1,121 @@
+from typing import List, Tuple
+
+from ..build_components.metadata import MaterialWithBuildMetadata
+from . import BaseMaterialAnalyzer
+
+
+class BasisMaterialAnalyzer(BaseMaterialAnalyzer):
+    """Analyzer for material basis (atomic positions and elements)."""
+
+    def get_layer_fingerprint(self, layer_thickness: float = 1.0) -> List[Tuple[Tuple[float, float], List[str]]]:
```
timurbazhirov, line 10: I think we should define fingerprint as a class - `LayeredFingerprintAlongAxis` to understand the data structure and allow multiple axes

src/py/mat3ra/made/tools/analyze/lattice.py
```diff
@@ -27,22 +29,48 @@ def material_with_conventional_lattice(self: MaterialWithBuildMetadata) -> Mater
...
     if original_number_of_atoms == primitive_structure_number_of_atoms:
-        # Not reduced, return original material if requested, to avoid unnecessary editions
         if return_original_if_not_reduced:
             return material
-    # Reduced, return the primitive structure
+
+    if keep_orientation:
+        primitive_analyzer = LatticeMaterialAnalyzer(material=material_with_primitive_lattice)
+        material_with_primitive_lattice = primitive_analyzer.get_material_with_corrected_orientation(material)
+        print("Orientation corrected after primitive conversion.")
```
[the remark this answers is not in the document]
VsevolodX: z -> -z, reversed oder.
I kept them separate to keep get_primitive a simple wrapper

src/py/mat3ra/made/tools/analyze/basis.py
```diff
@@ -0,0 +1,121 @@
...
+
+            if layer_elements:
+                unique_elements = sorted(list(set(layer_elements)))
+            else:
+                unique_elements = []  # Empty layer
+
+            layers.append(((layer_min, layer_max), unique_elements))
+            current_z += layer_thickness
+
+        return layers
```
timurbazhirov, line 50: We do have atomic layers object already - could we consider reusing that, btw?

src/py/mat3ra/made/tools/analyze/basis.py
```diff
@@ -0,0 +1,121 @@
...
+        # Compare fingerprints using similarity scores
+        normal_score = self._compare_fingerprints(original_fingerprint, current_fingerprint)
+        flipped_score = self._compare_fingerprints(original_fingerprint, list(reversed(current_fingerprint)))
+
+        # If flipped orientation has significantly higher similarity, material is flipped
+        # Use a threshold to determine if the difference is significant enough
+        threshold = 0.1  # 10% difference threshold
+        return flipped_score > normal_score + threshold
+
+    def _compare_fingerprints(
```
timurbazhirov, line 79: get_similarity_score - and can be a function of the `LayeredFingerprintAlongAxis` class

src/py/mat3ra/made/tools/analyze/basis.py
VsevolodX: It's a bit different, I think. This one gives a snapshots of element at each layer, not necessarily the physically separate layers. So it's an ordered  list, not an actual structure.

src/py/mat3ra/made/tools/analyze/interface/zsl.py
```diff
@@ -140,6 +150,54 @@ def convert_generated_match_to_match_holder(self, match_id: int, match_pymatgen)
...
+            substrate_supercell_matrix = substrate_supercell_matrix // g
+            area = area / g**2
+
+        return film_supercell_matrix, substrate_supercell_matrix, area
+
+    def _calculate_strain(self, film_sl_supercell_vectors, substrate_sl_supercell_vectors):
+        film_det = np.linalg.det(film_sl_supercell_vectors)
+        substrate_det = np.linalg.det(substrate_sl_supercell_vectors)
+
+        if (film_det > 0) != (substrate_det > 0):
```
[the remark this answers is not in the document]
VsevolodX: It's handling when lattices match, but are oriented in the opposite direction. So HEX can have two directions for one vector, different by 180 degrees, but being equivalent. The strain would still use 2 magnitudes for calulcation

timurbazhirov (approved):

tests/py/unit/test_tools_build_interface_zsl.py
```diff
@@ -87,11 +116,14 @@ def test_zsl_interface_builder(substrate, film, gap, vacuum, max_area, expected_
 
     builder = InterfaceBuilder()
     interface = builder.get_material(interface_config)
+    # interface = get_material_with_primitive_lattice(interface)
```
timurbazhirov, line 119: Remove?

PR #248, opened 2025-08-22
feature/SOF 7704

VsevolodX:
- **update: detect lattice type (gemini-cli)**
- **update: add test**
- **update: a map between pmg an ESSE**
- **update: adjust the map**


src/py/mat3ra/made/tools/analyze/lattice/analyzer.py
```diff
@@ -0,0 +1,73 @@
+from ....lattice import LatticeTypeEnum
+from ...build_components.metadata import MaterialWithBuildMetadata
+from ...convert import from_pymatgen, to_pymatgen
+from ...third_party import PymatgenSpacegroupAnalyzer
+from .. import BaseMaterialAnalyzer
+from .utils import detect_lattice_type_from_vectors
+
+PYMATGEN_LATTICE_TYPE_MAP = {
```
timurbazhirov, line 8: We should be able to identify other types too:

https://github.com/Exabyte-io/express/blob/5ded4f0da38a5e11e43161090b555ab90c09b482/express/parsers/structure.py#L129

VsevolodX: Let me check that, because it didn't get the FCC in tests

PR #270, opened 2026-04-02
update: add reciprocal lattice calcs

VsevolodX:


src/py/mat3ra/made/lattice.py
```diff
@@ -128,6 +128,17 @@ def cell_volume(self) -> float:
...
+    def reciprocal_vector_norms(self) -> List[float]:
+        reciprocal_vectors = np.linalg.inv(np.array(self.vector_arrays, dtype=float)).T
+        return [float(np.linalg.norm(vector)) for vector in reciprocal_vectors]
+
+    @property
+    def reciprocal_vector_ratios(self) -> List[float]:
+        norms = self.reciprocal_vector_norms
+        max_norm = max(norms)
+        return [round(float(value / max_norm), 3) for value in norms]
+
```
timurbazhirov, line 144: We should be able to get this from lattice.reciprocal_vectors, no?

VsevolodX: we need to add reciprocal_vectors property then, it's only present in JS

timurbazhirov (approved):

PR #276, opened 2026-06-08
add non periodic lattice defaults

VsevolodX:


src/py/mat3ra/made/tools/convert/utils.py
```diff
@@ -2,12 +2,54 @@
...
+    if len(shifted_positions) < 2:
+        return max(default_min_size, 10.0)
+    distances = pdist(shifted_positions)
+    max_distance = float(np.max(distances))
+    min_distance = float(np.min(distances))
+    padding_factor = _get_non_periodic_lattice_padding_factor(max_distance, min_distance)
+    return max(default_min_size, max_distance * padding_factor)
+
+
+def strip_poscar_comments(poscar_string: str) -> str:
```
timurbazhirov: We should use https://github.com/Exabyte-io/utils/blob/3a80497b01536ba4621b9abb583dd08c2e7864cb/src/py/mat3ra/utils/string.py#L62. Similarly for JS. Should be Fortran comments.

src/js/lattice/lattice.ts
```diff
@@ -22,6 +22,12 @@ import { UnitCell, UnitCellProps } from "./unit_cell";
  */
 export const nonPeriodicLatticeScalingFactor = 2.0;
 
+export const defaultNonPeriodicMinimumLatticeSize = 3.0;
```
timurbazhirov: We can keep this in JSON to reuse in Py

src/js/lattice/lattice.ts
```diff
@@ -22,6 +22,12 @@ import { UnitCell, UnitCellProps } from "./unit_cell";
...
+
+export const diatomicLatticePaddingFactor = 3.0;
+
+export const molecularLatticePaddingFactor = 2.0;
```
timurbazhirov: This is identical to nonPeriodicLatticeScalingFactor - or just put one equal to another

timurbazhirov (approved):

PR #288, opened 2026-07-04
chore/SOF-7951 Chore: use utils instead of code/utils

VsevolodX:
code/utils changed to utils
also JS uses banker's rounding like Python -- to resolved hashes for materials correctly


tests/fixtures/feo-standata.json
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
timurbazhirov: Let's use standata in tests?

src/js/basis/basis.ts
```diff
@@ -1,12 +1,14 @@
...
 import { InMemoryEntity } from "@mat3ra/code/dist/js/entity";
-import { math } from "@mat3ra/code/dist/js/math";
 import { BasisSchema, Coordinate3DSchema, Vector3DSchema } from "@mat3ra/esse/dist/js/types";
+import { Utils } from "@mat3ra/utils";
 import { chain, toPairs, uniq, values } from "lodash";
 
 import { Cell } from "../cell/cell";
 import { ATOMIC_COORD_UNITS, HASH_TOLERANCE } from "../constants";
+
+const { math } = Utils;
```
timurbazhirov: Elsewhere we usually use Utils.math inside the code instead of just "math" for clarity

timurbazhirov (changes requested):

src/js/basis/basis.ts
VsevolodX: I figured this would have less change -- just 2 lines, not many per file. 
But I can do that if that's how we want it

tests/fixtures/feo-standata.json
VsevolodX: Ok. I think I did that on other commit... Let me reapply

src/js/cell/primitive_cell.ts
```diff
@@ -1,5 +1,9 @@
-import { math } from "@mat3ra/code/dist/js/math";
 import { LatticeSchema, Matrix3X3Schema } from "@mat3ra/esse/dist/js/types";
+import { Utils } from "@mat3ra/utils";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const math: any = Utils.math;
```
timurbazhirov: Let's just use Utils.math to avoid having to redefine

timurbazhirov (approved):

